PR #3, opened 2015-09-16
Migrate multi step indicator

KorinaGraber:
Ported the multi-step indicator to components.

PR #4, opened 2015-09-16
Card container

KorinaGraber:
Started porting the card container over to components. Ported sorts, filters, and most of the data sources.

Dependent on https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/3


KorinaGraber:
Finished porting the card container to components

KorinaGraber:
This may require some testing, as I also converted all card container pieces to use bindToController in the process

PR #9, opened 2015-09-21
Added functionality for checking if the state is active when the indi…

KorinaGraber:
…cator initializes. This will show the correct state as highlighted when the user redirects to the state through a direct link.


SonofNun15:
Very nice!

PR #13, opened 2015-09-29
Convert to ES6 style external module syntax

SonofNun15:
Sorry, massive PR.

Converts all components to use the new ES6 external module syntax.


KorinaGraber:
:shipit:

PR #17, opened 2015-09-30
Include component dependencies

SonofNun15:
Aside from major shared libraries (angular, moment, jquery, lodash, typescript-angular-utilities), all component dependencies should be include in components so they don't have to be dealt with in consuming applications.


SonofNun15:
This could have some interesting ramifications in RL2... we'll probably have to remove any components that are already there so they are not redundant (the slider for example)

KorinaGraber:
:shipit:

PR #18, opened 2015-09-30
Remove textAngular

SonofNun15:
Removing textAngular from components for now. It doesn't play nicely with webpack.

We'll have to figure out a solution for this in MUSIC when we get there...


SonofNun15:
Didn't change the version. Wasn't thinking this had to get published immediately

KorinaGraber:
:shipit:

PR #27, opened 2015-10-07
Message log and dialog fixes

KorinaGraber:
Editable message log was not getting included in the angular module for components. Fixed this by including it directly in the message log module instead of using a sub-module.

Implemented functionality for resolving 'resolve' settings in base dialog and autosave dialog. Autosave dialog will append the results of the resolve actions to the data object, and they'll be included in the parameter used to call the autosave action.


KorinaGraber:
Requires utilities publish of https://github.com/RenovoSolutions/TypeScript-Angular-Utilities/pull/34

KorinaGraber:
Updated utilities

PR #34, opened 2015-10-12
Implemented functionality for applying the resolve data directly to t…

KorinaGraber:
…he dialog scope if the consumer doesn't want to specify a controller for simple use cases. This functionality is required by a resolveToDialog flag. In addition, the consumer may specify an alias for the data object using 'dialogAs', to follow common controllerAs practices


SonofNun15:
Nice

PR #35, opened 2015-10-13
Rich text editor

KorinaGraber:
Implemented a new rich text editor in components using ng-wig. ng-wig is dependency-free, which fixes all the headache we've had trying to resolve the dependencies for a rich text editor library, and it also seems much easier to customize than textAngular or any other option I've seen so far

image


SonofNun15:
Can we flip the order of the buttons? Then :shipit:

PR #39, opened 2015-10-15
Multi step spinner

KorinaGraber:
Implemented functionality for showing a spinner when executing a step's click event. This can be used either for the default click (state change) or for a consumer-specified click, if they return a promise. (If they don't, the spinner will just hide right away) Also disable further clicks until the promise resolves.


SonofNun15:
Looking good! :grinning:

PR #65, opened 2015-11-18
Select

KorinaGraber:
Created a component for select dropdowns. For now, this component supports:
required and custom validation
flower-up labels
asynchronous data loading (behavior is visually clunky right now)
selector for display name

Additional features can be added as we need them.

No selection:
image

Open (hides selection box): 
image

Selection with flowered-up label:
image

### DIFF
--- a/index.html
+++ b/index.html
@@ -19,7 +19,9 @@
 
 		<div ng-app="app">
 			<div ng-controller="TestController as test">
-				<rl-rich-text-editor ng-model="test.text"></rl-rich-text-editor>
+				Selection: {{test.selection}}
+				<rl-select ng-model="test.selection" options="test.options" option-as="item"
+						   selector="'name'" label="Test selector"></rl-select>
 				<rl-spinner ng-model="test.number"></rl-spinner>
 				<rl-date-time ng-model="test.date"></rl-date-time>
 			</div>
--- a/source/components/select/select.ts
+++ b/source/components/select/select.ts
@@ -97,7 +97,6 @@ export function select(): angular.IDirective {
 		bindToController: {
 			options: '=',
 			getOptions: '&',
-			optionAs: '@',
 			selector: '=',
 			validator: '=',
 			label: '@',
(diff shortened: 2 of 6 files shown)


SonofNun15:
You say "hides selection box". Are you saying that the "menu" / popover opens on the drop down instead of below it?

KorinaGraber:
Yes, that is the behavior of ui-select

index.html
```diff
@@ -19,7 +19,9 @@
 
 		<div ng-app="app">
 			<div ng-controller="TestController as test">
-				<rl-rich-text-editor ng-model="test.text"></rl-rich-text-editor>
+				Selection: {{test.selection}}
+				<rl-select ng-model="test.selection" options="test.options" option-as="item"
```
SonofNun15, line 23: Ended up realizing that the messageLog templating has an issue with this currently... Since the ng-repeat uses an query expression, it doesn't currently support `message-as`. Didn't realize this was an issue at first because my templates must have been too simple. But looks like you have to assume that `$entry` will be available as the message for now. Can't see a way around it at the moment... :disappointed:

source/components/select/select.ts
```diff
@@ -0,0 +1,111 @@
...
+		restrict: 'E',
+		require: 'ngModel',
+		template: require('./select.html'),
+		controller: controllerName,
+		controllerAs: 'select',
+		scope: {},
+		bindToController: {
+			options: '=',
+			getOptions: '&',
+			optionAs: '@',
```
SonofNun15: Doesn't look like `optionAs` is currently used / implemented?

KorinaGraber: True. It's really only useful if you have custom templating. I was figuring we could leave that as an option for later, in case someone wants to show a checkmark or something like that in the dropdown list

index.html
KorinaGraber, line 23: Yeah, you would almost need to use a directive to alias the item in the ng-repeat, sort of like what we do with card container cards

source/components/select/select.ts
SonofNun15: Agreed

SonofNun15: Should probably drop for now and we can implement as a rl-custom-select or something

SonofNun15: Because it will be more complicated

PR #69, opened 2015-11-19
Validation group

KorinaGraber:
Created a component for validating against a group of components. This component will show an error message at the top if the group has a validation error. It also hooks into the angular-form system so we can track errors above it.

Dependent on https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/68


SonofNun15:
Merge conflict

PR #73, opened 2015-11-24
Fix for generic container

SonofNun15:
Configured templates were getting replaced by templates loaded from the DOM.

Instead we want to merge the templates.


KorinaGraber:
:shipit:

PR #74, opened 2015-11-25
Autosave refinements

KorinaGraber:
Moved autosave service and autosave action service to components. Implemented functionality for specifying triggers for the autosave. Default trigger is 'onChange', which triggers debounced autosave on value change. Autosave dialog disables this by setting the trigger to 'none' (Add ability to set the triggers in settings)


KorinaGraber:
Added ability to specify the triggers for an autosave dialog. Defaults to 'none' since the dialog triggers a save on close.

PR #85, opened 2015-12-11
Card container builder - first pass

KorinaGraber:
First pass implementation of a card container builder. For now in order to be backwards compatible, the card container just accepts a builder in addition to the other bindings and copies the builder properties to its controller.


KorinaGraber:
Updated the interface and moved the logic for setting card container properties to the builder. Not sure if you're still saying it would be good to allow for useClickableCards(false) or not. We could always add that later, I guess

PR #91, opened 2015-12-16
Allow null option

KorinaGraber:
Added support for specifying a null option for the select dropdown. When a null option is specified, add a config object to the start of the options array. The code will recognize this object and use the nullOption string as its display name, and set ngModel to null when it's selected. This isn't a pretty solution, but it's the only thing I could come up with due to the fact that ui-select doesn't support null-options yet.

Per https://github.com/angular-ui/ui-select/issues/302, null options aren't supported in ui-select yet.

**TODO**: Add unit tests for the select component. This could come in a later PR. I tested this feature visually to make sure that it worked.

### DIFF
--- a/index.html
+++ b/index.html
@@ -21,7 +21,7 @@
 			<div ng-controller="TestController as test">
 				Selection: {{test.selection}}
 				<rl-select ng-model="test.selection" options="test.options" option-as="item"
-						   selector="'name'" label="Test selector"></rl-select>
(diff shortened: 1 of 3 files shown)


SonofNun15:
ui-select doesn't support null options? That seems kind of lame.

SonofNun15:
Does Select2? I thought ui-select was based on Select2?

KorinaGraber:
As per the linked issue, they're thinking of adding support, but haven't yet. I don't know about Select2

PR #92, opened 2015-12-17
Generic Select Filter

JoshDSommer:
Created a component for adding select filters to the card container.


KorinaGraber:
It's a little confusing that the consumer has to specify a valueSelector to the filter service and a display name selector to the filter component. Unfortunately, I'm not sure if there's a way around that. Both are needed.

SonofNun15:
@SamGraber: You reference setting the display name in your comment above. Can you link to where this is done? Don't recall seeing the display name part. Thanks!

KorinaGraber:
That would be the 'selector' for the rl-select component.
https://github.com/TheOriginalJosh/TypeScript-Angular-Components/commit/da620bb38021962288ff3ded05b97ad0cc152505#diff-eacf331f0ffc35d4b482f1d15a887d3bR44

PR #95, opened 2015-12-18
Filter group enhancements

KorinaGraber:
Hide count from the IFilterOption interface. This is an internal implementation detail for the filter group. Also make type optional. If not specified, the label is used as the type.

Should probably be a minor version, as this will be a breaking change for @TheOriginalJosh


SonofNun15:
Let's discuss

PR #96, opened 2015-12-23
Styling updates

KorinaGraber:
Tweaks to component styling to match theme updates. Brent is still working on the field validation styling a little bit, but this should get the html a little closer to what it needs to be. 

Fixed multi-step errors and validation groups.

### DIFF
--- a/source/components/busy/busy.ts
+++ b/source/components/busy/busy.ts
@@ -9,7 +9,7 @@ function busy(): angular.IDirective {
 	'use strict';
 	return {
 		restrict: 'E',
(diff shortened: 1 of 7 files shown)


SonofNun15:
This whole PR is very specific to a proprietary style sheet. We need to revisit and consider how these changes may affect the components in general for other use cases and other styling.

SonofNun15:
What direction are we going with this PR?

KorinaGraber:
Needs some design/discussion with Brent

SonofNun15:
I guess if we want to add default styling for the busy component we can add that in the next PR.

PR #101, opened 2016-01-04
Allow message data to be put on the scope

SonofNun15:
This lets templates access stuff from the parent scope. This is a hack, but is the only easy way to pass stuff from outer parent scopes through to the message scope.


KorinaGraber:
:shipit:

PR #103, opened 2016-01-11
Fix generic container

KorinaGraber:
After some investigation, I determined that any new elements need to be added to the parent _before_ they are compiled for them to be able to correctly access the parent context. For the generic container, that means that we need to resolve the templates to JQuery objects in the template loader. Then in the swap templates function, we can replace the container content with the template element, and then compile it. This also means we'll need to map code-generated templates to JQuery elements on init.

Stack overflow: http://stackoverflow.com/a/27645175/4220059


SonofNun15:
Nice fix, very clean

PR #108, opened 2016-01-18
If the result is not a promise, hide the spinner unless the action re…

KorinaGraber:
…turns true. (Essentially, the action needs to return true to specify that a spinner should be shown)


SonofNun15:
Is this a breaking compatibility change?

KorinaGraber:
Yes, probably so

KorinaGraber:
Changes the behavior so you now have to return 'true' explicitly if you want a spinner to stay indefinitely

SonofNun15:
OK, merging but we'll increment the minor version on next release

PR #110, opened 2016-01-19
Columns removing inline style

BrentWMiller:
Moving inline style from component to Theme project. The style will be applied to `.template-container` instead of being an inline style.

https://github.com/RenovoSolutions/RenovoTheme/pull/84


SonofNun15:
Waiting on new publish of theme version

PR #111, opened 2016-01-20
Date filter

JoshDSommer:
New date filter for card container.


SonofNun15:
Had a number of comments, but solid PR on the whole. Nice work, thanks for putting this together!

PR #115, opened 2016-01-27
Added the ability to specify ng-disabled for the rich text editor. Th…

KorinaGraber:
…is doesn't do anything yet, but it will once the changes to support ng-disabled in ng-wig come through.


SonofNun15:
Any comments or movement on that PR on the ngWig repo?

KorinaGraber:
Nope, not that I've seen

PR #121, opened 2016-02-09
updates for Service Event List server side filtering and sorting. 

JoshDSommer:
added build buildCustomDataSource, for server filters/sorts of  the card containter. added fieldName for storing the field name for sortings. made the sort direction value public for server sorting


SonofNun15:
I confused. You removed the work in progress label, but don't see any additional commits... ? :confused:

JoshDSommer:
No additional changes this is so that the card container doesn't filter on the front end when using the custome data source.

PR #124, opened 2016-02-10
Dialog component

KorinaGraber:
Created a component for specfying the content of a dialog. At this point, the component just transcludes in a header, content and footer. The important thing here is that it abstracts away the bootstrap modal classes, which makes it really easy to migrate to an alternative dialog type. Also, if the consumer sets the 'autosave' option, the footer defaults to the autosave dialog footer.


KorinaGraber:
Also, this would make it easy to add an 'x' and have it universally affect all of our dialogs.

SonofNun15:
Curious to see how this get consumed... Basically this is going to be given to the dialog service when opening a dialog?

SonofNun15:
Looks like you can basically still give the modal service whatever you want, but in general we'll use this component. I wish we could bake more of this into the actual service, but that's tough since you still have to specify the header, body, footer somehow. Looks cool, thanks for putting this together!

KorinaGraber:
:grinning: You're welcome. I pretty much did it more or less for fun and to de-stress with something fun after dealing with different issues.

PR #127, opened 2016-02-16
Server side data source

KorinaGraber:
Implemented a server-side data source. This data source serializes the values from the filters and sends them along with the sorts and paging details to the server for processing. This is similar to the old implementation but no longer requires building a complex data model for the filters on the server.

**This is a breaking change due to the fact that serverSearch data source was renamed to clientServerDataSource**

**Requires utilities >= 2.7.6** https://github.com/RenovoSolutions/TypeScript-Angular-Utilities/pull/90


KorinaGraber:
Updated mode and range filter groups to serialize to `null` if the selected option has no value. (displayAll for mode filter)

SonofNun15:
Merge conflict

KorinaGraber:
Resolved merge conflicts

PR #128, opened 2016-02-16
Message Log - Layout Fixes for Remove Button [Markup]

BrentWMiller:
# Message Log - Layout Fixes for Remove Button [Markup]

This PR goes along with [RenovoTheme PR #108](https://github.com/RenovoSolutions/RenovoTheme/pull/108). This will break a message log message down into pieces that can be styled against. We can place the 'remove' button properly without any kind of overlapping in elements.

msg-log-delete-gray


SonofNun15:
Looks great, nice work.

PR #135, opened 2016-02-19
Use builder

KorinaGraber:
Use the card container builder in the header and footer components of the card container.

**This was still under discussion as to whether we want to implement it this way since this will prohibit use of the card container without using the builder**


SonofNun15:
Still not crazy about this, but I know it does reduce our complexity by eliminating the container service...

Mainly just seems like a violation for a component to rely on it's builder for behavior. I would expect the component to stand / work alone and the builder (by name / function) to just assemble the component.

PR #136, opened 2016-02-19
Actually, we don't want to send the paging to the server. Instead spe…

KorinaGraber:
…cify a throttle limit as the page size. We'll just set the page number to 1 for now.


KorinaGraber:
Test failures

KorinaGraber:
Added build fixes

PR #143, opened 2016-02-25
Read only datetime

JoshDSommer:
addition of rlDateTimeStatic for disiplaying dates with the timezone, currently hardcoded EST awaiting UTC date update.


KorinaGraber:
:shipit:

PR #144, opened 2016-02-25
Smart data source fixes

KorinaGraber:
Fixes for the smart data source:
- Need to force the source to reload on init. (Initial refresh won't trigger a reload if throttle is false) 
- Added smart data source to the card container builder. 
- Trim all null filters from the applied filters. I thought the toDictionary was already going to do this, but realized afterwards that this just trims out all filters without a label.


SonofNun15:
Merge conflict

SonofNun15:
:shipit:

PR #148, opened 2016-03-01
Popover

KorinaGraber:
Created a behavior to wrap angular-ui popovers. In Angular-ui 0.14 they dropped support for binding in a template without using a templateUrl, which doesn't work very well for us since we use require() with webpack. The new popover expects a binding that evaluates to an html template. Or with text-only set to true, a template can be specified inline for simple text templates.


SonofNun15:
Merge conflict

PR #151, opened 2016-03-02
Altered the dialog close button to use our new styles. This will stac…

BrentWMiller:
…k two FontAwesome icons on top of each other so that they can be individually animated and styled. I moved the click event to the surrounding span so it triggers no matter where you click on the button.

[RenovoTheme PR #112](https://github.com/RenovoSolutions/RenovoTheme/pull/112)

modal-close-final


KorinaGraber:
:shipit:

PR #113, opened 2016-01-26
Going to Lodash 4

SonofNun15:
**Requires**: https://github.com/RenovoSolutions/TypeScript-Angular-Utilities/pull/81

PR #152, opened 2016-03-02
Message Log Fixes

BrentWMiller:
- Changed **button type** to match changes being made in the Theme. It doesn't work how it is right now anyway since `btn-` is added to the front, making it `btn-message-delete-button`.
- Added a ng-class to apply a `.system-note` class. This means we can style system notes and user notes differently.

Grabs styling from [RenovoTheme PR #113](https://github.com/RenovoSolutions/RenovoTheme/pull/113)


KorinaGraber:
:shipit:

PR #156, opened 2016-03-03
rl-date-time-static component upgrade angular 1.5

JoshDSommer:


KorinaGraber:
Probably okay to :shipit:. Most of the comments were suggestions for future improvements, rather than necessarily things to change right away

PR #157, opened 2016-03-03
Typeahead

KorinaGraber:
Created a new typeahead component using ui-select instead of angular-ui. The new component simplifies the behavior significantly. The typeahead will now be used on one of two ways:
- Consumer wants to select a single value. In this case, the consumer can enable the 'collapse' feature, which will cause the typeahead to collapse and show the value when a selection is made. In addition, you can allow for selecting the search value in this case, in case the consumer wants the user to be able to create a manual entry.
- Consumer wants to add selections to a list. For this, the consumer simply needs to provide a select handler. When a selection is made, the handler will get called, and the consumer can add the item to his list.

Additional considerations:
As of right now, the component doesn't show any kind of spinner. I'll want to add support for that shortly.
I would like to add support for validation next, as in the other input components.

https://renovo.myjetbrains.com/youtrack/issue/IDEAS-6


KorinaGraber:
Added a spinner

KorinaGraber:
Added custom validation support.

KorinaGraber:
**Breaking change**

KorinaGraber:
Merged in https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/158
**Requires utilities >= 2.9.1** https://github.com/RenovoSolutions/TypeScript-Angular-Utilities/pull/98

PR #158, opened 2016-03-04
Transform

KorinaGraber:
Use the new transform helper in utilities to update all usages of transform to support a function or string. **This is a breaking change** because in some cases, the transform was getting bound in as an event binding, while now it needs to be a one-way or two-way binding.

Breaking changes:
Some card container interfaces now need a generic type.
Comma list now uses two-way binding for 'transform', not event binding.

**Requires utilities >= 2.9.1** https://github.com/RenovoSolutions/TypeScript-Angular-Utilities/pull/98


SonofNun15:
:shipit:

PR #159, opened 2016-03-07
Optional bindings

KorinaGraber:
Use one-time optional bindings for select. This is breaking when some of these bindings are missing. The only thing that is truly required for select is ng-model, although there won't be anything to pick unless options OR get-options is specified.


SonofNun15:
Merge conflict

SonofNun15:
:shipit:

PR #160, opened 2016-03-07
Template renderer

KorinaGraber:
Created a template renderer component. Takes an object with a template and a scope and renders it in the view. Added an option for specifying that templates should be rendered within the card container. If true, every filter with a 'template' string will get rendered beside the card container. The filters get rendered with 'filter' and 'dataSource' available on the scope.

https://renovo.myjetbrains.com/youtrack/issue/IDEAS-1


source/components/cardContainer/filters/cardContainerFilters.ts
```diff
@@ -0,0 +1,66 @@
...
+	renderableFilters: IRenderableFilter[];
+
+	static $inject: string[] = ['$rootScope'];
+	constructor(private $rootScope: angular.IRootScopeService) { }
+
+	$onInit(): void {
+		this.renderableFilters = _(this.filters).filter((filter: IRenderableFilter): boolean => {
+			return filter.template != null;
+		}).map((filter: IRenderableFilter): IRenderableFilter => {
+			let scope: IFilterScope = <IFilterScope>this.$rootScope.$new();
```
SonofNun15, line 42: Creating this new scope from the root scope could be an interesting issue, but it's hard to tell what else could really be used here. Could be passed in from somewhere externally, or could derive from the scope of this directive.

PR #161, opened 2016-03-07
Filter templates

KorinaGraber:
Added templates for the dateFilter and selectFilter. Requires specifying the component options on the service itself, since we can no longer bind to these.

Requires https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/158, https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/157, and https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/160


SonofNun15:
_Another_ conflict

PR #163, opened 2016-03-07
Multi transclusion

KorinaGraber:
Use multi-transclusion wherever possible to simplify transclusion in our components. It appears there may be a way to call the transclude function to get only a specific transclude slot, which would be game changing for the way the card container works. My concern is that this isn't meant to be a publicly accessible API, and therefore they may change the interface. 
Issue: https://github.com/angular/angular.js/issues/14193


KorinaGraber:
Confirmed that this is a documented feature: https://code.angularjs.org/1.5.0/docs/api/ng/service/$compile#-controller-

SonofNun15:
:shipit:

PR #164, opened 2016-03-07
Header alias

KorinaGraber:
Implemented support for applying the cardAs alias to card header columns. If the user specifies a custom template for the column, they should be able to access the item using the same alias.


KorinaGraber:
https://renovo.myjetbrains.com/youtrack/issue/INF-118

SonofNun15:
:shipit:

PR #169, opened 2016-03-09
Fix transclusion

KorinaGraber:
If the slot isn't filled, the transclude function doesn't even get called. We need to check for if the slot is filled. If not, compiled the default header/footer by hand.


jtest15:
Most likely pulled into RL2 in https://github.com/RenovoSolutions/renovoLive2/pull/2002

jtest15:
https://renovo.myjetbrains.com/youtrack/issue/RLv2-1621

PR #180, opened 2016-03-17
Ability to specify card type

BrentWMiller:
This will allow us to specify a card type on the simpleCard component using `card-type=""`.


source/components/simpleCardList/simpleCard.ts
```diff
@@ -79,6 +80,8 @@ export class SimpleCardController implements ISimpleCardBindings {
 				this.close();
 			}
 		});
+
+		this.cardType = this.cardType != null ? this.cardType : null;
```
KorinaGraber: FYI - you actually don't need this. Angular will automatically set the cardType via a binding. This was just if you wanted to have a default value

BrentWMiller: Oh awesome, that makes it even more simple.

KorinaGraber:
:shipit:

PR #181, opened 2016-03-18
Angular validation

KorinaGraber:
Tie into the angular form system for validation This allows us to get away from having to define so much of our own custom 'aggregate' validation logic to handle the same thing. We'll use the component validator service to tie validation errors to the ngModel, so we can pull them off the form when we want to show a validation error. 

I apologize for the size of this PR. There was a lot that went into it.

**Requires utilities >= 2.9.3** https://github.com/RenovoSolutions/TypeScript-Angular-Utilities/pull/103

Other changes
- Created a new rl-required directive for required errors with a message.
- Created a base class for inputs to share the validation logic.
- Combined the dialog and autosaveDialog services. An autosaving dialog is simply opened using dialog.openForm. (An autosaving dialog is essentially just a dialog with a form that we want to preserve the contents of)
- Fixed an issue where the autosave dialog wasn't picking up the form. Instead we'll just include a form in the dialog component. 
- onChangeTrigger now watches $dirty and $valid and only saves when both are true. 
- Added functionality for setting saveWhenInvalid to specify that the form should save even when $valid  for cases where you want to have work-in-progress validation. (Also affects the onChangeTrigger)
- Created an rl-form component to wrap an angular form with an attached save function. When the form is submitted, validate against the angular form model and call the save function if validation passes.
- Changed autosave form dialogs to use a spinner on the save button instead of closing and showing the autosave bar


source/components/textarea/textarea.html
```diff
@@ -1,5 +1,5 @@
 <div class="field" ng-class="{ 'error': input.ngModel.$invalid }">
 	<label ng-show="input.inputValue && input.label" class="show-hide angular-animate">{{input.label}}</label>
-	<textarea class="form-control" ng-model="input.inputValue" name="{{input.name}}" rows="{{input.rows}}" maxlength="{{input.maxlength}}" ng-disabled="input.ngDisabled" placeholder="{{input.label}}"></textarea>
+	<textarea class="form-control" ng-model="input.inputValue" name="{{input.name}}" maxlength="{{input.maxlength}}" ng-disabled="input.ngDisabled" placeholder="{{input.label}}"></textarea>
```
SonofNun15: Accidentally removed the `rows=` property, not the `name=` property.

SonofNun15: https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/181/commits/3e80ba354949c08bb28435b3e2cf087716bd15cb

source/services/dialog/dialog.service.ts
```diff
@@ -58,6 +46,50 @@ export class DialogService<TDialogSettings> implements IDialogService<TDialogSet
 
 		return this.dialog.prompt(options, require('./promptDialog.html'));
 	}
+
+	openAutosaveForm(options: IAutosaveDialogSettings): void {
+		this.promise.resolvePromises(options.resolve).then((resolveData: any): void => {
+			let scope: IAutosaveDialogScope = <IAutosaveDialogScope>options.scope;
+
+			if (scope == null) {
+				scope = <IAutosaveDialogScope>this.$rootScope.$new();
```
SonofNun15: Do we want to base this on root scope, or should we consider the $parent scope as the parent? Probably can wait until we see how this is used and / or how it needs to be used.

SonofNun15: Nevermind, we don't have access to any scope in this context, so can't create a scope from the parent. By parent, I meant the calling context, but there is no way to do this.

source/components/form/form.ts
```diff
@@ -0,0 +1,73 @@
...
+	constructor(private $element: angular.IAugmentedJQuery
+			, private $scope: IFormScope
+			, private $timeout: angular.ITimeoutService
+			, private $q: angular.IQService
+			, private autosaveFactory: IAutosaveServiceFactory) { }
+
+	$onInit(): void {
+		this.$element.find('form').on('submit', (): void => {
+			this.autosave.validateAndSave();
+			this.$scope.$apply();
```
SonofNun15: It doesn't seem like this is going to prevent the actual form from submitting, which could result in some unexpected behavior.

SonofNun15: Need to investigate if `ng-submit` is required, or `$event.preventDefault()`

SonofNun15: Per @SamGraber, probably don't want to `preventDefault` because angular still needs to hear the event as well.

PR #183, opened 2016-03-22
Adding inject to dateTimeStatic componet

JoshDSommer:
Broken bindings on the service even list page.
https://renovo.myjetbrains.com/youtrack/issue/MUSIC-354


KorinaGraber:
That would do it

PR #184, opened 2016-03-22
Adding .field & changing group error class name

BrentWMiller:
Added `.field` to our **dateTime** and **spinner** components.
- Will need some small fixes in the Theme.

I also changed `content-group` error messages to use the class name `.error-message` instead of just `.error`. This will make sure their styling is scoped better and not overwriting other element's classes.
- This change will require a Theme PR to change styling from `.error` to `.error-message`. Should just be an easy name switch.
#### This will all still work - just ugly without theme updates.

https://github.com/RenovoSolutions/RenovoTheme/pull/124

### DIFF
--- a/source/components/spinner/spinner.html
+++ b/source/components/spinner/spinner.html
@@ -1,6 +1,6 @@
 <rl-generic-container selector="spinner.ngDisabled">
 	<template default>
-		<div ng-class="{ 'error': spinner.ngModel.$invalid }">
+		<div class="field" ng-class="{ 'error': spinner.ngModel.$invalid }">
 			<input name="{{spinner.name}}" class="spinner" id="{{spinner.spinnerId}}" type="text" />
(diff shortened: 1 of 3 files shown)


source/components/spinner/spinner.html
```diff
@@ -1,6 +1,6 @@
 <rl-generic-container selector="spinner.ngDisabled">
 	<template default>
-		<div ng-class="{ 'error': spinner.ngModel.$invalid }">
+		<div class="field" ng-class="{ 'error': spinner.ngModel.$invalid }">
```
KorinaGraber, line 3: Do you also need to set this on the disabled spinner below? I suppose validation styling in particular isn't very relevant if the spinner is disabled

BrentWMiller, line 3: Added it just to be safe.

KorinaGraber, line 3: Cool

KorinaGraber:
:shipit:

PR #185, opened 2016-03-22
Performance tuning buttons

KorinaGraber:
Use one-time bindings for all watches that should only get set when the component is initialized. This primarily includes the classes and whether the spinner should be on the right or left. (Still need an active watch for when the spinner shows)

Optimized to:
Button          1
ButtonAsync     6
ButtonLink      2
ButtonSubmit        6 (1 more with 'saving')
ButtonToggle        5

Previously:
Button          4
ButtonAsync     13
ButtonLink      7
ButtonSubmit        13
ButtonToggle        8


KorinaGraber:
ButtonLink now has 2 watches again, as we want the link to be dynamically updatable from outside.

PR #191, opened 2016-03-25
Documentation

KorinaGraber:
Documented async buttons, lazy load, rl-form, rl-required, rl-autosave, and autosave triggers.


BrentWMiller:
Awesome!

PR #194, opened 2016-03-29
Fix typeahead link

KorinaGraber:
Need to link the typehead list to the typeahead so it can keep the two lists properly in sync.


KorinaGraber:
Ran `All`

PR #195, opened 2016-03-29
Cleanup

KorinaGraber:
Updated most of the components to use component syntax. Cleaned up the sample markup to group the controls using headers and labels.

image


KorinaGraber:
Dropped in a simple ui-router:
image

PR #197, opened 2016-04-01
Prevent default attribute

JoshDSommer:
I had created this to address an issue with bootstrap's collapse feature. however it is not being used now. Regardless it may come in handy in the future... maybe?


KorinaGraber:
Actually, we already have this. Use `rl-stop-event-propagation="myEvent"`. https://github.com/RenovoSolutions/TypeScript-Angular-Utilities/blob/master/source/behaviors/stopEventPropagation/stopEventPropagation.ts

JoshDSommer:
LOL well, I shall cancel this.

PR #199, opened 2016-04-04
Card container update

JoshDSommer:
Update to allow setting of additional classes on the card containter header columns. to help with https://renovo.myjetbrains.com/youtrack/issue/MUSIC-426

### DIFF
--- a/source/components/cardContainer/card/headerColumn/headerColumn.ts
+++ b/source/components/cardContainer/card/headerColumn/headerColumn.ts
@@ -33,7 +33,7 @@ export class HeaderColumnController {
 		$scope.$on('card.refresh', this.update); //*event?
 	}
 
-	private update: {(): void} = (): void => {
+	private update: { (): void } = (): void => {
 		this.value = __transform.getValue(this.item, this.column.getValue);
 	}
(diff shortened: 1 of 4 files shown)


SonofNun15:
This is a bit of an odd solution, but I guess it is workable. Styling is definitely not a breakpoint size, so this definitely breaks down semantically. I know you spent a lot of time looking into this, so I'm going to assume this was a compromise for the sake of actually getting it to work.

JoshDSommer:
Correct, I believe the attribute directive that sets this was causing the issue. This was the work around that seemed to work the best. this is an example column with the builder and I think it works pretty well

 `{
            label: "<span class=\"hidden-md hidden-lg\">Pr</span><span  class=\"hidden-sm hidden-xs\">Priority</span>",
            size:
            {
                xs: 0,
                sm: 1,
                styling: 'no-padding text-center'
            },
            getValue(item: serviceEventList.ServiceEventListItem): number {
                return item.priority;
            },
            template: require('./serviceEventColumn/serviceEventColumnPriority.html'),
            fieldName: 'Priority'
        }`

PR #201, opened 2016-04-05
Time zones

KorinaGraber:
dateTime and dateTimeStatic should expect a moment object with a timezone. Show the timezone on the date-time input or to the right of the static date. For dateTime, use formatters and parsers to expose the date to the input as a string but reconstruct it as a moment when a new date is selected. The selected timezone set as the timezone on the moment. This gives us the ability to allow the user to change the time zone if we want to provide that functionality at a later date.

Updated dateTime and spinner to inherit from the InputController for shared behavior. They still use their own directive definitions. It appears that we actually can require to controller for directive definitions, so I'm not sure what the issue was for previous times when it didn't seem to work.


KorinaGraber:
All components outside of the card container now use component syntax.

KorinaGraber:
Ran `all`

source/components/dateTimeStatic/dateTimeStatic.ts
```diff
@@ -24,20 +24,21 @@ export interface IDateTimeStaticController extends IDateTimeStaticBindings {
...
 	includeTime: boolean;
 	displayValue: string;
 	displayTimeZone: boolean;
 
 	static $inject: string[] = [__date.serviceName];
 	constructor(private dateUtility: __date.IDateUtility) {
-		this.displayValue = '';
 		if (this.dateValue != null && this.dateUtility.isDate(this.dateValue)) {
-			this.displayValue = moment(this.dateValue).format(__date.defaultFormats.dateFormat);
+			this.dateValue = moment(this.dateValue);
```
SonofNun15, line 37: Before this line, what are you expecting `this.dateValue` to be? Is it OK to call `moment()` on a `moment.Moment` type?

source/components/dateTimeStatic/dateTimeStatic.ts
```diff
@@ -24,20 +24,21 @@ export interface IDateTimeStaticController extends IDateTimeStaticBindings {
...
+
+			this.displayValue = this.dateValue.format(__date.defaultFormats.dateFormat);
```
SonofNun15: This is redundant with line 41, so perhaps it should just move into an else for perf reasons?

source/components/dateTimeStatic/dateTimeStatic.ts
KorinaGraber, line 37: Yes it is. moment() seems to function like $q.when() in that it's a way to verify that it's a moment.

source/components/dateTimeStatic/dateTimeStatic.ts
SonofNun15: Nevermind, looking more closely I see that 41 is composed with this assignment.

KorinaGraber: Updated anyway. The if here will just format using the dateTimeFormat.

source/components/dateTime/dateTime.ts
```diff
@@ -159,7 +159,16 @@ function dateTime(moment: moment.MomentStatic
 			let max: string | Date | moment.Moment
 				= dateTime.max != null ? dateTime.max : defaults.maxDate;
 
-			scope.$watch((): any => { return ngModel.$viewValue; }, (newValue: any): void => {
+			ngModel.$formatters.push((value: moment.Moment): string => {
+				return moment(value).format(getFormatOrDefault());
```
SonofNun15: Again: interesting that you are calling `moment(value)` on an object that, by the typings, appears to already be a `moment` object... ?

SonofNun15:
:shipit:

PR #205, opened 2016-04-06
Date filter

KorinaGraber:
Created an `rlDate` filter for handling static dates. This is much more concise then using the dateTimeStatic component. Brent doesn't think we'll need to style the timezone here, since it's just appended to the end of the string.

```
<rl-date-time-static date-value="myDate" include-time="true"></rl-date-time-static>
{{myDate | rlDate:true}}
```


KorinaGraber:
Ran `all`

PR #208, opened 2016-04-11
Docs and example cleanup

KorinaGraber:
Documented checkbox, commaList, date filter, dialog component, and genericContainer. Added links to the documentation from the bootstrapper. Cleaned up the examples so they look better.


KorinaGraber:
Added some performance tuning for radio buttons
Previous:

```
Radio               3
RadioGroup          3
```

Tuned:

```
Radio               1
RadioGroup          2
```

source/components/genericContainer/genericContainer.md
```diff
@@ -0,0 +1,44 @@
...
+
+#### `when-selector`
+Specifies the value that the container should match against to determine if this template should be shown. Unlike `ng-switch`, interpolation is permitted.
+
+#### `default (default: false)`
+If specified, the specified template is used if no match is found. In this case, no `when-selector` need be specified.
+
+Template contents must be surrounded with an html tag to function correctly.
+
+### Full example
```
SonofNun15, line 24: Might be worth clarifying why the generic container is different and / or superior to ng-switch. Basically because you can have it resolve templates based on expressions.

KorinaGraber, line 24: That was the intention of this line: https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/208/files#diff-26c6de6951c135a78992633aede9c813R2. I can say more on the subject if you think it's necessary.

PR #213, opened 2016-04-14
Added a webpack configuration and converted the bootstrapper to Typescript

KorinaGraber:
Added the ability to webpack the bootstrapper to enable typescript and bundle all the bootstrapper files into a single file. Converted the bootstrap controllers to Typescript and moved them into individual subfolders for reorganization.


JoshDSommer:
:shipit:

PR #214, opened 2016-04-14
Address exception being thrown when datetime filter is selected. left…

JoshDSommer:
… commented out code for reference.


KorinaGraber:
:shipit:

PR #216, opened 2016-04-14
Switching to `require` for the moment. `import` seems to break the bu…

KorinaGraber:
…ild for html files for some reason. I think because tsc can't understand it.


SonofNun15:
Spot on. TypeScript doesn't know how to import *.html files, so until we can use something like SystemJS with TS compiling, we'll have to stick with `require` for html templates and other assets.

KorinaGraber:
I thought I asked you recently if we could import other types of files with webpack, and you said we could. Oh well.

SonofNun15:
Well, at the time I thought that it would work because it works with SystemJS when transpiling TS. But it's the TSC compiling that it breaks on.

KorinaGraber:
Right. Yeah, I noticed that.

PR #217, opened 2016-04-14
Bootstrap issues, eliminate cardSearch watches.

KorinaGraber:
As much as we can, it would be nice to eliminate watches in our components in particular, since these will have the biggest performance impact on the application. We've had some problems with properties not playing nicely with angular, but I feel like it still makes sense to use them in cases where they function correctly.

Updated the bootstrapper to bundle the library along with the bootstrap application instead of having a script reference. This saves us from having to bundle both when changes are made.


KorinaGraber:
Drop usages of $watch in favor of usage of $onChanges lifecycle hook, observables, and TS properties  where applicable. TS properties sometimes don't play nicely with angular, so we have to be cautious in our usage of them.

source/components/simpleCardList/simpleCard.ts
```diff
@@ -136,6 +136,7 @@ export class SimpleCardController implements ISimpleCardBindings {
 	private updateAlwaysOpen(alwaysOpen: boolean): void {
 		if (alwaysOpen) {
 			this.showContent = true;
+			this.alwaysOpen = true;
```
SonofNun15, line 139: Angular will set this normally, so may want to just update this in the test before calling $onChanges instead.

KorinaGraber:
Oops, pushed the new changes straight upstream.

SonofNun15:
That works I guess... :smirk:

PR #219, opened 2016-04-18
Add class "spinner-container" to spinner containers

ibedard16:
Done to resolve https://renovo.myjetbrains.com/youtrack/issue/MUSIC-457 .
Will not do anything without also merging this pr: https://github.com/RenovoSolutions/RenovoTheme/pull/143


KorinaGraber:
Actually, I don't think you should need this. From what I was seeing the SE list page wasn't using our spinner components, which _doesn't_ show the duplicate spinner. So you should just need to update the date filter to use the spinner component instead of its own spinner.

PR #224, opened 2016-04-21
Update to address multiple types in buttons

JoshDSommer:
Also addressed an issue where if a size wasn't being set a `btn-` was still appending to class attribute.

### DIFF
--- a/bootstrapper/buttons/buttons.html
+++ b/bootstrapper/buttons/buttons.html
@@ -28,7 +28,7 @@
 		<a href="https://github.com/SamGraber/TypeScript-Angular-Components/tree/master/source/components/buttonToggle/buttonToggle.md">rl-button-toggle</a>:
 	</div>
 	<div class="col-xs-11">
-		<rl-button-toggle ng-model="button.toggled" on-toggle="button.action('Toggle button')">Toggle button</rl-button-toggle>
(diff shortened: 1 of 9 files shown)


SFarageNIS:
This is the size of PR's we like to see! :+1:

KorinaGraber:
Can we see an example usage of this in the bootstrapper? 

Also, I think ButtonToggle _does_ inherit from the base button, so we shouldn't need that duplicate types function. 
https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/224/commits/21d989bd9b559765e06e15be985ac4e19be97237#diff-8d5757bc1c8b862e111cbfec047b8c83R37

https://github.com/TheOriginalJosh/TypeScript-Angular-Components/blob/update_to_address_multiple_types_in_buttons/source/components/buttonToggle/buttonToggle.ts#L24

JoshDSommer:
your right the button toggle does inherit from the base button.It seems kind of excessive to need it in the bootstrapper since we don't have examples for all the other attrs. but here you go.

KorinaGraber:
Yeah, it probably isn't necessary long term. Maybe a commit note showing an example usage in one of the commits? Could be useful for future PR's like this anyway. The main thing is it took me a bit to sort through what the types helper was doing. Looks good though!

PR #225, opened 2016-04-21
Typeahead list fixes

KorinaGraber:
Set the typeahead ngModel to dirty when items are added/removed. Hide the header when the list is empty.

Also fixed issue with genericContainer where $onChanges was getting triggered before the container was set up. The initial template will always be set after initialization, so we don't need to watch for changes until after the setup.


KorinaGraber:

PR #231, opened 2016-04-29
Bootstrapper improvements

KorinaGraber:
Made each bootstrapper component its own module. This removes the need for a separate bootstrapApp file. Also moved the route config for each bootstrapper component into its own file. Changed the base navigation to a left nav. If necessary, we can do nested navigation along the top for bigger component sections.

image


KorinaGraber:
Merging, since this only touches the bootstrapper.

PR #232, opened 2016-04-29
Use rxjs observables

KorinaGraber:
Installed rxjs and cherry picked the commit to migrate to rxjs for angular 2. Converted data sources to use streams instead of the old observables. 

WIP - build fails on some typings issue in rxjs. Also, could migrate additional components before merging this.


KorinaGraber:
Finished except for simple card list, the usage of which I'm unsure of how to convert to a stream. May need some investigation or help with someone more familiar with rxjs streams.

KorinaGraber:
Also need to run the tests.

KorinaGraber:
Fixed tests.

KorinaGraber:
Finished removing usages of the observable service in components.

package.json
```diff
@@ -89,6 +89,7 @@
     "moment-timezone": "^0.5.3",
     "ng-wig": "~2.3.4",
     "rx": "~4.0.8",
```
SonofNun15, line 91: You may have already done this, but ultimately we can remove the reference to `rx`

source/components/cardContainer/dataSources/smartDataSource/smartDataSource.service.ts
```diff
@@ -125,7 +125,7 @@ export class SmartDataSource<TDataType> extends AsyncDataSource<TDataType> {
 		this.subscriptions = [];
 		_.each(this.filters, (filter: filters.ISerializableFilter<any>): void => {
 			if (_.isFunction(filter.subscribe)) {
-				this.subscriptions.push(filter.subscribe((): void => { this.onFilterChange(filter); }));
+				this.subscriptions.push(<any>filter.subscribe((): void => { this.onFilterChange(filter); }));
```
SonofNun15, line 126: Doesn't filter.subscribe return a Subscription now?

package.json
KorinaGraber, line 91: I didn't do that yet in this PR. I was trying to be cautious about where it might conflict with changes in the Angular 2 migration

source/components/cardContainer/dataSources/smartDataSource/smartDataSource.service.ts
KorinaGraber, line 126: I'm not sure. It might now

KorinaGraber, line 126: I think as-of the pre-Angular 2 version of utilities, the subscribe function still returns a `Subscriber` on the interface

PR #233, opened 2016-05-02
Localize dates

JoshDSommer:
this is a filter for taking the UTC dates in a string and localizing them based on the current timezone.

### DIFF
--- a/source/filters/localizeStringDates/localizeStringDates.filter.ts
+++ b/source/filters/localizeStringDates/localizeStringDates.filter.ts
@@ -1,0 +1,67 @@
+'use strict';
+
+import * as angular from 'angular';
+import * as moment from 'moment';
+
+import { services } from 'typescript-angular-utilities';
+import __date = services.date;
+import __object = services.object;
+import __timezones = services.timezone;
+
+
+export var moduleName: string = 'rl21.ui.filters.localizeStringDates';
+export var filterName: string = 'rlLocalizeStringDates';
+
+export interface ILocalizeStringDates {
+	(input?: string): string;
+}
+
+/**
+ *Filters a string and find all, datetimes that match the format M/D/YYYY H:M:S AM|PM that is used to store dates in message log body text.
+* Assumes they are UTC  and parses them, converts to local browser time replaces them in the string.
+ */
+export function localizeStringDates(): ILocalizeStringDates {
+	'use strict';
+	return (input?: string): string => {
+		if (input == null) {
+			return '';
+		}
+
+		//if no timezone is specified then use moment to get the browsers time zone.
+		const timezone: string = __timezones.timezoneService.currentTimezone.momentName;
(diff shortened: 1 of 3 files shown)


source/filters/localizeStringDates/localizeStringDates.filter.ts
```diff
@@ -0,0 +1,67 @@
...
+export function localizeStringDates(): ILocalizeStringDates {
+	'use strict';
+	return (input?: string): string => {
+		if (input == null) {
+			return '';
+		}
+
+		//if no timezone is specified then use moment to get the browsers time zone.
+		const timezone: string = __timezones.timezoneService.currentTimezone.momentName;
+		const regex: RegExp = /([1-9]|1[0-2])\/([1-9]|1\d|2\d|3[01])\/(19|20)\d{2} ([1-9]|1[0-9])\:([0-9])([0-9])(?:\:([0-9])([0-9]))? (AM|PM)(?: UTC)?/g;
```
KorinaGraber, line 34: Is there a way you could break this up into named components and then combine them to make it more readable?
Such as:

```
const hoursExp = /([1-9|1[0-2];
...
const regex = hoursExp + minutesExp + ...
```

Not a huge fan of big 'greek' regular expressions. (Basically all regular expressions bigger than 10 characters or so)

KorinaGraber:
Tests look good

source/filters/localizeStringDates/localizeStringDates.filter.ts
SonofNun15, line 34:
```
const hoursExpPattern: string = '([1-9|1[0-2])';
const otherPattern: string = '([1-9][1\d|2\d|3[01])';
const pattern: string = [hoursExpPattern, otherPattern].join('/');
const regex: RegExp = new RegExp(pattern);
```

SonofNun15, line 34: Could work

JoshDSommer, line 34: I fail to see how that is less complicated looking. the nice thing about it be a large greek one is anyone can take it an copy and past it into an online editor with minimal effor such as https://regex101.com/r/eU1fD5/1

KorinaGraber, line 34: The main advantage would just be splitting up the different segments of the expression into named variables explaining their purpose, such as 'years', 'months', 'days', 'hours', 'minutes', etc, so someone can see at a glance what's going on. I basically look at that expression and then quickly look away, hoping that it does what it's supposed to. But I suppose being able to copy and paste it into regex101.com is an advantage.

JoshDSommer, line 34: I see what your getting at. How about I leave a comment instead? that would be less overall code but also would allow anyone else working on it in the future the ability to know at a glance.

KorinaGraber, line 34: Sure, if you can find a way to explain what each piece of the expression is doing

SFarageNIS, line 34: Steven strongly agrees with Sam/SonOfNun about breaking out the RegEx into component names.

PR #237, opened 2016-05-09
Select fixes

KorinaGraber:
Added the ability to specify a custom select item template in the select filter. Unfortunately, this requires a complicated hack of transcluding in the template, then binding it to the select component. I'm hoping this will be easier in angular 2. 

Fixed a bug where using a custom template would cause the null option to show incorrectly.


JoshDSommer:
:shipit:

PR #238, opened 2016-05-09
Use a behavior subject instead of having a separate subscribe functio…

KorinaGraber:
…n to implement the same behavior.


JoshDSommer:
:shipit:

PR #239, opened 2016-05-09
Need to provide a default 'empty' jquery element.

KorinaGraber:
Because we call `.filter` on it, which breaks if it's not specified.


SFarageNIS:
But why... :)

KorinaGraber:
Haha, sorry. I'm not used to that yet.

PR #241, opened 2016-05-10
Typeahead list - hide dropdown if empty

KorinaGraber:
When searching is disabled, hide the dropdown if there are no options left. In this case there is no action that is possible with the dropdown, so it makes more sense to hide it altogether.


JoshDSommer:
:shipit:

PR #245, opened 2016-05-17
Standardizing 'transform' terminology

KorinaGraber:
Always use `transform` as the binding name to keep things standardized. This will make it much easier to remember how to use various components. For starters, changed `select` and `selectFilter` to use `transform` instead of `selector`.


KorinaGraber:
Added a root bootstrap app component to make it easier to reuse the bootstrapper elsewhere. Also added a serve script that can be run to start `http-server` from VSCode (With the npm integration). I still like having http-server running separately from the build sometimes in case an error kills the build and it has to be restarted.

PR #246, opened 2016-05-19
Documentation

KorinaGraber:
Documented stringWithWatermark, validationGroup, and userRating.

### DIFF
--- a/bootstrapper/forms/formsBootstrapper.ts
+++ b/bootstrapper/forms/formsBootstrapper.ts
@@ -1,12 +1,24 @@
 import * as angular from 'angular';
 
+import { services } from 'typescript-angular-utilities';
+import __validation = services.validation;
+
 export const moduleName: string = 'FormTestModule';
 
 class FormTestController {
 	count: number;
+	text: string;
+	validator: __validation.IValidationHandler;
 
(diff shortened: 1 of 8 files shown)


bootstrapper/forms/formsBootstrapper.ts
```diff
@@ -1,13 +1,25 @@
...
 	count: number;
+	text: string;
+	validator: __validation.IValidationHandler;
 
 	static $inject: string[] = ['$q', '$timeout'];
 	constructor(private $q: angular.IQService, private $timeout: angular.ITimeoutService) {}
 
+	$onInit(): void {
+		this.validator = {
+			validate: () => this.text === 'valid',
```
JoshDSommer, line 18: Do we care about magic strings in the bootstrappers?

KorinaGraber, line 18: That's a fair question. The purpose of the bootstrapper is different, since it's designed to demonstrate and test functionality. We could certainly change this to some other validator, like string length > 5-10 or something.

SFarageNIS:
@TheOriginalJosh if this needs additional work, let's talk about it some more.  don't want PR's getting old if we can avoid it.

PR #248, opened 2016-05-27
Add initialize as dirty to form and dialog to allow (optional) saving…

arknotts:
… a form with no changes.

https://renovo.myjetbrains.com/youtrack/issue/MUSIC-546


JoshDSommer:
:shipit:

PR #251, opened 2016-06-06
Fix custom search feature and clean up timing issues

JamieBowen:
Removed the hasSearchOption variable.  The first custom option was being added correctly, but usage of the hasSearchOption flag was preventing subsequent custom options from being added.

Changed the return type of the create property from void to any.  This matches the interface definition.

Removed $q.when wrapper on call to this.getItems(), as the getItems function is defined to return a promise.  Wrapping a promise in $q.when is an anti-pattern.

Refactored branch of loadItems function that handles the useClientSearching case.  When useClientSearching is enabled, the API should be called one time to load the entire list, then the search text is filtered on the client.  However, it was possible for a user who types quickly to trigger the API call more than once.  Fixed this by caching the promise returned from the getItems function.

### DIFF
--- a/source/components/typeahead/typeahead.ts
+++ b/source/components/typeahead/typeahead.ts
@@ -106,7 +106,7 @@ export class TypeaheadController extends InputController {
 	childLink: __parentChild.IChild<ITypeaheadBehavior>;
 	hasSelection: boolean;
 	select: { (params: ISelectParams): void };
-	create: { (params: ICreateParams): void };
+	create: { (params: ICreateParams): any };
 	transform: { (item: any): string } | string;
 	getItems: { (params?: IGetItemsParams): angular.IPromise<any> };
 	prefix: string;
@@ -122,7 +122,6 @@ export class TypeaheadController extends InputController {
 	collapseOnSelect: boolean;
 	allowCustomOption: boolean;
 	collapsed: boolean = false;
-	hasSearchOption: boolean = false;
 
 	get selection(): any {
 		return this.ngModel.$viewValue;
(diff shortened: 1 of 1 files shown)


JamieBowen:
Resolves issue: https://renovo.myjetbrains.com/youtrack/issue/MUSIC-580

source/components/typeahead/typeahead.ts
```diff
@@ -215,29 +214,41 @@ export class TypeaheadController extends InputController {
...
 			if (this.showCustomSearch(search)) {
-				this.hasSearchOption = true;
 				this.visibleItems.unshift(this._searchOption);
 			}
 		});
 	}
 
 	loadItems(search: string): angular.IPromise<void> {
 		if (!this.useClientSearching) {
-			return this.$q.when(this.getItems({
```
JamieBowen: Refactored for readability

source/components/typeahead/typeahead.ts
```diff
@@ -215,29 +214,41 @@ export class TypeaheadController extends InputController {
...
 
 			if (this.showCustomSearch(search)) {
-				this.hasSearchOption = true;
 				this.visibleItems.unshift(this._searchOption);
 			}
 		});
 	}
 
 	loadItems(search: string): angular.IPromise<void> {
 		if (!this.useClientSearching) {
```
JamieBowen: Refactored for readability

source/components/typeahead/typeahead.ts
```diff
@@ -215,29 +214,41 @@ export class TypeaheadController extends InputController {
...
-			return this.$q.when(this.getItems({
-				search: search,
-			})).then((items: any[]): void => {
-				this.visibleItems = items;
-			});
```
JamieBowen: Refactored for readability

JoshDSommer:
:shipit:

PR #253, opened 2016-06-08
Check validation result before searching

JamieBowen:
The data source was being refreshed and bound regardless of whether calling validateSearchLength resulted in a search length error or not.  This resulted in the UI showing a validation error but still performing the search.

Also removed unused import IDataSource.

Unit tests passed.

Resolves issue: https://renovo.myjetbrains.com/youtrack/issue/MUSIC-532


JoshDSommer:
nice :shipit:

PR #255, opened 2016-06-10
Cleanup

KorinaGraber:
Made some changes from the discussion. Select elements from the template instead of injecting the element. Remove custom exception and rely on the angular DI exception. Added typings for module for module.id.
Unrelated - renamed longClickButton to buttonLongClick.ng1. It's nice to keep button as the prefix so the folder will appear next to the other buttons.


KorinaGraber:
Implemented long click button in angular 2

KorinaGraber:
Revert back to using require() for html files. We'll consider looking into using the angular offline compiler to resolve this later.

PR #256, opened 2016-06-13
Show/hide UI message w/o API call

JamieBowen:
These changes help resolve an issue that was causing the message log to continue displaying an uploading graphic after a failed upload.

The call to dataService.saveMessage was resolving immediately due to a promise anti-pattern inside the Attachments Service.  This caused an API call to be made via getTopPage before the new message was actually saved.  It did, however, incidentally cause the message being added to show up on the screen because the Attachments Service was temporarily adding it to the list when getMessages was called (again via getTopPage).

In a separate commit, the Attachments Service was updated to correctly return the actual promise for the file upload.  Subsequently, getTopPage was no longer being executed right away, and it made more sense for the MessageLog Service to temporarily add the message to the visibleMessages itself without making that API call through the Attachments Service.  Then the MessageLog Service also needs to remove the new message on failure, thus the need to import the arrayUtility.

### DIFF
--- a/source/components/messageLog/messageLog.service.ts
+++ b/source/components/messageLog/messageLog.service.ts
@@ -1,5 +1,3 @@
-'use strict';
-
 import * as ng from 'angular';
 import * as moment from 'moment';
 
--- a/source/components/messageLog/messageLog.service.ts
+++ b/source/components/messageLog/messageLog.service.ts
@@ -1,5 +1,10 @@
+'use strict';
+
 import * as ng from 'angular';
 import * as moment from 'moment';
+
+import { services } from 'typescript-angular-utilities';
+import __array = services.array;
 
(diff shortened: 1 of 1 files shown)


JamieBowen:
Resolves issue: https://renovo.myjetbrains.com/youtrack/issue/MUSIC-597#

source/components/messageLog/messageLog.service.ts
```diff
@@ -1,6 +1,11 @@
+'use strict';
```
KorinaGraber: FYI - we're no longer using `'use strict'` at the top of our files. This is a detail that typescript handles for us now, so we shouldn't need to worry about it anymore

KorinaGraber: I'm going to remove it in the core-services change since I already merged that

PR #258, opened 2016-06-14
Button and input subfolders

KorinaGraber:
Created subfolders for buttons and inputs. For now only the components with Angular 2 versions have been moved. This is nice for organization, and also allows us to put an index file at the root that serves as a 'bucket' for the angular 2 components in these folders. Also, we can compile a list of directives of these types, for ease of use.


SFarageNIS:
Yea for less magic!  Nice improvement and organization!

PR #260, opened 2016-06-16
Card container filters

KorinaGraber:
Removed angular specific code from filter services. We'll only make these available through the card container builder for now. Added angular 2 components for the filters alongside the angular 1 versions.
Implemented SerializableFilter in dateFilter and selectFilter.

For now the dateFilter only supports going back in weeks. What would be nice is if we could roll our own spinner component, and then add the ability to specify a dynamic addon. That way we could use the spinner with the days/weeks toggle.


KorinaGraber:
This is a minor breaking change for filters, since the factory is no longer injectable

PR #262, opened 2016-06-16
Card container data sources

KorinaGraber:
Remove angular specific code from the data source classes. For now these will all be required to be instantiated through the card container builder.

Built on top of https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/260


KorinaGraber:
Also converted the sorter services

PR #264, opened 2016-06-21
Tests and card cleanup

KorinaGraber:
Added unit tests for comma list, form, form service, and component validator. Implemented string with watermark in angular 2.

Converted both card list types to use @ children decorators instead of registering a list of cards. This lets us depend on angular to ensure that only the visible cards will appear in the list, and that they are in the correct order. Since the order is now dependable, we can apply a class to the 'odd' cards in a simple card list.

Requires https://github.com/RenovoSolutions/TypeScript-Angular-Utilities/pull/138

### DIFF
--- a/bootstrapper/misc/miscNg2.html
+++ b/bootstrapper/misc/miscNg2.html
@@ -13,6 +13,12 @@
 </div>
 
 <div class="form-group">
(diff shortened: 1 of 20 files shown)


JoshDSommer:
:shipit:

PR #266, opened 2016-06-23
Reorganization

KorinaGraber:
Typeahead and typeahead list get renamed to .ng1 and moved to the inputs folder. buttons and inputs should be available on the library tree so consumers can use the ng2 version.

The v2.1.2 commit snuck in here too. This version was already published to npm


KorinaGraber:
Merging as this is mostly just reorganization. @SFarageNIS can review later if desired. The time pipe may change or go away, since we're considering handling times differently

PR #269, opened 2016-06-27
Absolute time component

KorinaGraber:
Created a component for selecting absolute times. The component uses a string in 'h:mmA' format at its data model. For internal use, it parses it into a time data model with hour, minute, and period. ('AM'/'PM') The period is shown as a toggle to the right of the input, while clicking the input opens a dual hour/minute dropdown. When an hour/minute is selected, the hours/minutes collapse to a selection. If the other is already selected, the popout automatically collapses for convenience. A selected hour or minute can be clicked to reopen the full list of options.

As yet, manual entry is not supported


SFarageNIS:
Pair reviewed.  :shipit:

PR #271, opened 2016-06-30
Validation group

KorinaGraber:
Implemented a validation group in angular 2. Because of the way change detection works in angular 2, you now have to pass in a model for it to watch on. Alternately, you can call `checkValidity()` to manually trigger the validation to reevaluate. Unfortunately, using a model doesn't always work with the way we use angular 2 right now, because the change detection only first if the model instance changes. Since we're not using immutable objects yet, this doesn't always work as it should.


SFarageNIS:
Paired :shipit: (on Tuesday!)

PR #273, opened 2016-07-01
Page size control was missing.

BrentWMiller:
The ng1 version needed to pull the matching ng1 template.


KorinaGraber:
Good catch

PR #276, opened 2016-07-06
Simulate keyup event in datepicker

JamieBowen:
Clicking the minute up/down buttons in the datepicker control was flooding the message log.  The changes were not being debounced because the autosave directive only listens for keyup events.  Simulating a keyup event in the datepicker's change event handler triggers the debouncing features.

Resolves: https://renovo.myjetbrains.com/youtrack/issue/MUSIC-624


JoshDSommer:
:shipit:

KorinaGraber:
Nice!

PR #277, opened 2016-07-06
When returning from the form submit, only return a boolean

KorinaGraber:
All the consumer needs to know is whether the form successfully initiated the submission. The actual wait value is only relevant for showing the spinner and marking the form as pristine. This solves a dilemma where the wait value needs to be something other than false to avoid ambiguity between a false wait value and validation failures.


SFarageNIS:
Good improvement.

PR #282, opened 2016-07-08
Upgrade wrapper services

JamieBowen:
Resolves issue: https://renovo.myjetbrains.com/youtrack/issue/RLv21-34

### DIFF
--- a/source/componentsDowngrade.ts
+++ b/source/componentsDowngrade.ts
@@ -29,6 +29,7 @@ import { MergeSort } from './components/cardContainer/sorts/mergeSort/mergeSort.
 import { DatePipe } from './pipes/date/date.pipe';
 
 import { FormService } from './services/form/form.service';
+import { WindowService } from './services/windowWrapper/windowWrapper.service';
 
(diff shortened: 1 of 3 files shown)


source/componentsDowngrade.ts
```diff
@@ -67,6 +69,7 @@ export function downgradeComponentsToAngular1(upgradeAdapter: UpgradeAdapter) {
 
 	upgradeAdapter.addProvider(DEFAULT_THEME_PROVIDER);
 	upgradeAdapter.addProvider(FormService);
+	upgradeAdapter.addProvider(WindowService);
```
KorinaGraber, line 75: There are actually two steps to downgrading an angular 2 provider:
1: Add the provider to the upgrade adapter. This makes it available for anyone to inject and essentially takes the place of providing it on the ng2 app (which we don't have)
2: Register the downgraded provider to angular 1, via:
`componentsDowngradeModule.factory(windowServiceName, upgradeAdapter.downgradeNg2Provider(WindowService));`
The first one is something we will be doing for all of our singleton services until we get rid of Angular 2, which is why the form service is adding itself as a provider here. (I was confused about what you were asking on Slack)

You'll have to complete step 2 for the window and document services. Sorry I wasn't more clear.

PR #284, opened 2016-07-08
Angular 2 dialog

KorinaGraber:
Created components for opening dialogs and prompt dialogs in angular 2. Because angular 2 doesn't have the concept of turning strings into templates, we can't use a service anymore. Instead, the consumer will need to put a dialog in their template, and programmatically tell it to open. The dialog itself will still open in a global location on the page, so there is no concern about events bubbling up to the wrong places. Also, since the dialog templates can just use the consumer's concept, there no longer needs to be any concept of resolving to the dialog or passing data through. For this reason I opted not to implement any pre-loading when the dialog opens. Consumers will have to do that themselves.

Form/autosave behavior:
The dialog inherits from form, like simple card and card. If autosave is turned on and no footer is specified, the footer defaults to a cancel and save button, where the save button submits the form and closes the dialog upon completion. This function is available on the dialog API in case the consumer wants to call it programmatically or override the dialog footer.

If autosave is off, the consumer can specify an onClosing handler that catches close events and returns a boolean indicating if the close can proceed.

If autosave is on, any onClosing handler is ignored. If the form is pristine, the dialog closes without saving. If the dialog is dirty, the dialog validates and tries to submit, but won't close if there are validation errors. The user can shortcut these by 'dismissing' using the cancel or 'x' button, which prevents any attempt to submit.

Also, if the user clicks the save button (Or the consumer calls submitAndClose programmatically), the dialog will try to save and close _regardless_ of the dirty state of the form. This should remove the need for an initializeFormDirty property on the angular 2 dialog component.


KorinaGraber:
Implemented unit tests

KorinaGraber:
TODO: Providers

KorinaGraber:
Done

KorinaGraber:
Will need an outlet in AppCore before usage

SFarageNIS:
Paired :shipit:

PR #286, opened 2016-07-08
throw an exception if the form is invalid but there are no errors to …

KorinaGraber:
…show.

This will make it easier to track down validation edge cases, such as the one Isaac and I fixed recently


SFarageNIS:
What's a case where the form is "invalid"?  What does that mean here?  Is it our fault or the 'users'/input fault?

KorinaGraber:
In this case it's generally our fault. The purpose of this is to try to catch bugs in validation and make them more visible

PR #289, opened 2016-07-11
Upgrading breakpoint service

JoshDSommer:


source/componentsDowngrade.ts
```diff
@@ -66,7 +68,8 @@ export function downgradeComponentsToAngular1(upgradeAdapter: UpgradeAdapter) {
 	});
 
 	upgradeAdapter.addProvider(DEFAULT_THEME_PROVIDER);
-	upgradeAdapter.addProvider(FormService);
+	upgradeAdapter.addProvider(FormService); /// todo emulate for breakpoint.
```
KorinaGraber: I don't think we need this comment anymore, do we?

source/services/breakpoints/breakpoints.service.tests.ts
```diff
@@ -20,14 +18,14 @@ interface IVisibleBreakpointsMock {
...
-	var breakpoints: IBreakpointService;
+	let breakpoints: IBreakpointService;
 
-	var visibleBreakpoint: string;
-	var triggerResize: { (): void };
+	let visibleBreakpoint: string;
+	let triggerResize: { (): void };
 
 	beforeEach((): void => {
 		angular.mock.module(moduleName);
```
KorinaGraber: I don't think we want or need this anymore. Also, we shouldn't need to import the test module from utilities

PR #290, opened 2016-07-12
Angular 2 Typeahead

KorinaGraber:
Implemented typeahead in angular 2. Like select, we're using our own custom dropdown list here. (Using the same classes) However, instead of the 'input' being just a toggle for the list, it's actually a textbox here. The list generally opens when a search opens, however, at the moment a completed search can still be opened/closed like a dropdown. (i.e., if a search was entered, you can click on the input again to reopen the same result list)

### DIFF
--- a/source/components/inputs/typeahead/typeahead.html
+++ b/source/components/inputs/typeahead/typeahead.html
@@ -1,7 +1,7 @@
 <div class="field"
 	 [class.error]="!control.valid"
 	 (offClick)="close()">
-	<label class="label-slide angular-animate" [hidden]="!(search && label)">{{label}}</label>
+	<label class="label-slide angular-animate" [hidden]="hideFlowerup">{{label}}</label>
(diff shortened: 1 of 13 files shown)


source/components/inputs/typeahead/typeahead.html
```diff
@@ -0,0 +1,36 @@
+<div class="field"
```
BrentWMiller: If you add the `.typeahead` class here it will swap out the drop-down caret for a search icon.

image

KorinaGraber: 👍

KorinaGraber: https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/290/commits/13d3d99b6d1592294de412af5cac958fb1d2cebd

JamieBowen:
Pair reviewed :shipit:

PR #292, opened 2016-07-13
Migrate tab and tabset to Angular 2

arknotts:
This migrates the tab and tabset components to Angular 2. A couple things to note:
- rlTabHeader needed to be its own component (not just a <ng-content> placeholder) in order to grab its HTML and bubble it up to the tabset component for display.
- rlTab doesn't extend FormComponent, but rather just looks so see if any FormComponent exists within its ContentChildren. Any form that exists in there and has validation will add the error class to the header of the tab.
- A new MockQueryList<T> class was written to mock the ContentChildren object. This may be useful in the future to pull out into a shared library for reusing unless Angular implements a better way.

https://renovo.myjetbrains.com/youtrack/issue/RLv21-28

Pair reviewed with @TheOriginalJosh


source/components/tabs/tab.ts
```diff
@@ -1,16 +1,36 @@
...
 
 	constructor() {
+	}
+}
+
+@Component({
+	selector: 'rlTabHeader',
+	template: '<ng-content></ng-content>',
+})
+export class TabHeaderComponent {
```
KorinaGraber, line 11: Rather than getting the element ref and pulling the html out of it, we would like to use angular template refs to handle this. That allows consumers to have full access to their context when specifying tab content. The only drawback is that the selector has to become an attribute to allow for being on a template element (or using an asterisk), but it seems worth the tradeoff.
Examples:
https://github.com/RenovoSolutions/TypeScript-Angular-Components/blob/master/source/components/cards/cardHeader.template.ts
Used:
https://github.com/RenovoSolutions/TypeScript-Angular-Components/blob/master/source/components/simpleCardList/simpleCard.ts
and 
https://github.com/RenovoSolutions/TypeScript-Angular-Components/blob/master/source/components/simpleCardList/simpleCard.html#L6
For consistency, I think we should add template directives for all three (header, content, and footer). I can pair with you on this sometime if that would be helpful

KorinaGraber:
Nice job all around!

source/components/tabs/tab.ts
arknotts, line 11: OK let me take a look at those. It felt pretty dirty doing it that way but wasn't aware of the templates.

KorinaGraber:
Merge conflicts now. Probably from the tabs bootstrapper

arknotts:
Trivial merge conflicts from https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/293, which are now resolved.

KorinaGraber:
👍

arknotts:
Resolved additional merge conflict @SamGraber

PR #293, opened 2016-07-14
Upgrade MultiStepIndicator to Angular 2

JamieBowen:
Resolves: https://renovo.myjetbrains.com/youtrack/issue/RLv21-25

Paired with @TheOriginalJosh on bootstrapper work

### DIFF
--- a/source/components/multiStepIndicator/multiStepIndicator.html
+++ b/source/components/multiStepIndicator/multiStepIndicator.html
@@ -9,7 +9,7 @@
 			<div class="wrap">
 				<p class="badge" *ngIf="step.count">{{step.count()}}</p>
 				<p class="error" *ngIf="!step.getValid()"></p>
-				<p class="title">{{step.title}} <rl-busy [loading]="step.isLoading"></rl-busy></p>
+				<p class="title">{{step.title}} <rlBusy [loading]="step.isLoading"></rlBusy></p>
 				<p class="subtitle">{{step.subtitle}}</p>
 			</div>
 		</li>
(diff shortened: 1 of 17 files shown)


source/components/multiStepIndicator/multiStepIndicator.html
```diff
@@ -1,11 +1,15 @@
...
+			[class.completed]="step.getCompleted()"
+			[class.current]="step.isCurrent"
+			[class.active]="step.isActive && !anyLoading()">
 			<div class="wrap">
-				<p class="badge" ng-show="step.hasCount">{{step.count()}}</p>
-				<p class="error" ng-if="!step.getValid()"></p>
-				<p class="title">{{step.title}} <rl-busy loading="step.loading"></rl-busy></p>
+				<p class="badge" *ngIf="step.count">{{step.count()}}</p>
+				<p class="error" *ngIf="!step.getValid()"></p>
+				<p class="title">{{step.title}} <rl-busy [loading]="step.isLoading"></rl-busy></p>
```
KorinaGraber: <rl-busy> is not a native element, so you'll have to import and use our busy component in angular 2.

```
import { BusyComponent } from '../busy/busy';

// your component definition
@Component({
    directives: [BusyComponent],
})

// template
<rlBusy></rlBusy>
or
<rlBusy [loading]="true"></rlBusy>
```

Because Angular 2 handles component interactions well, we also decided to add a `trigger` API to the busy component. So instead of managing a loading flag at all, you could just pass the click promise to the `trigger` method and it'll spin until the promise completes. Although, this might not work very well since each step has its own busy.

```
@ViewChild(BusyComponent) busy: BusyComponent;

// in an action method
let myPromse = doSomething();
this.busy.trigger(myPromise); // spins until promise resolves
```

This could be done with a child component for the individual steps, but that would take more work. Otherwise it's fine as-is.

JamieBowen: Imported the `BusyComponent`... but I'll stick with managing the `loading` flag just because the `step.isLoading` property needs to stick around for use in the `anyLoading` function.

KorinaGraber: 👍

KorinaGraber:
:shipit:

PR #294, opened 2016-07-14
Dropdown list accessibility

KorinaGraber:
Implementing the ability to use the keyboard to select an item from a select or typeahead dropdown.

Per @JamieBowen's suggestion, the user can use the up/down arrows to scroll through the options, but the focus remains on the input. In fact, the focus now determines whether the list should show/hide. (For select, always show/hide on focus/blur. For typeahead, show if focused and searched) If the user hits enter or tabs away when an entry is highlighted, that entry is selected.

Built on https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/290
Starts at the `Typeahead accessibility - first pass` commit, but there are a few that overlap after that


JamieBowen:
Pair reviewed :shipit:

PR #296, opened 2016-07-14
Typeahead fixes

KorinaGraber:
Random improvements and fixes from the pair review with Jamie.

Also, I ran into a really tricky bug where the blur event suppresses the click on the popout items, which causes the items to not be clickable at all.
Instead of using a click event on the popout item, use a mouseenter and mouseleave event to set or clear the current focus item. That way when a blur happens on click, the highlighted item will still be selected.


KorinaGraber:
This is related: http://stackoverflow.com/a/10653160

JamieBowen:
:shipit:

PR #297, opened 2016-07-15
Update bootstrapper click events for MSI

JamieBowen:
Running setTimeout in the click event allows dev to visually test that the spinner is working.

Not sure what's up with the .md file line endings.

### DIFF
--- a/angular2Migration.md
+++ b/angular2Migration.md
@@ -1,65 +1,65 @@
 ###Component
-Busy - Done
-Button - Done
-ButtonAsync - Done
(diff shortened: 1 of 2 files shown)


KorinaGraber:
Merge conflict. Probably on the angularMigration md file. :shipit:

KorinaGraber:
Actually, this change to the migration md file won't work. We need two spaces at the end of each line to force the markdown viewer to newline there. This is why I don't use vscode for markdown... 😞

PR #298, opened 2016-07-15
Replace finally with then and catch

JamieBowen:
ES6 Promises do not support `finally` so the check for `isFunction(result.finally)` was failing and causing the spinner to not show.

Resolves: https://renovo.myjetbrains.com/youtrack/issue/MUSIC-702


JamieBowen:
Tracked and discussed the solution with @SamGraber

KorinaGraber:
:shipit:

PR #301, opened 2016-07-18
Tidy up promise rejections

JamieBowen:
Making sure errors do not get swallowed in MSI and buttonAsync.  Promise rejections need to be caught so the spinners can be stopped.  The errors then need to be rethrown, otherwise the promise chain will resolve from that point forward.  Unit tests were updated to ensure that the error is rethrown.


KorinaGraber:
Nice! :shipit:

PR #302, opened 2016-07-19
Typeahead list

KorinaGraber:
Implemented a typeahead list in angular 2

Manages the relationship between the list and the typeahead.

### DIFF
--- a/bootstrapper/forms/formsNg2Bootstrapper.ts
+++ b/bootstrapper/forms/formsNg2Bootstrapper.ts
@@ -7,7 +7,7 @@ import { services } from 'typescript-angular-utilities';
 import __timezone = services.timezone;
 
 import { FormComponent } from '../../source/components/form/form';
-import { ButtonSubmitComponent } from '../../source/components/buttons/buttonSubmit/buttonSubmit';
+import { BUTTON_DIRECTIVES } from '../../source/components/buttons/index';
 import { INPUT_DIRECTIVES } from '../../source/components/inputs/index';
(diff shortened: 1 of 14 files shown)


bootstrapper/forms/formsNg2Bootstrapper.ts
```diff
@@ -42,6 +42,7 @@ export class FormsBootstrapper {
 	options: ITestItem[];
 	optionsAsync: Observable<ITestItem[]>;
```
JoshDSommer, line 43: There is a naming convention I've seen others using and started using my self where observables are suffixed with `$` so in this case it would be `option$` I like this since it makes spotting observables very easy. probably something we should discus as a team

source/components/inputs/typeaheadList/typeaheadList.ts
```diff
@@ -0,0 +1,180 @@
...
+				this.cachedItems.next(this.cachedItemsArray);
+			}
+			return newItem;
+		});
+		return action;
+	}
+
+	remove = (item: T): Observable<void> => {
+		const action = this.asObservable(this.onRemove(item));
+		action.subscribe(() => {
```
JoshDSommer, line 105: I don't quite follow this, not saying it's bad or anything id just like to understand it better. would you want to pair on this at some point?

KorinaGraber, line 105: Sure, we could do that. I usually pair review these bigger PRs with someone anyway

PR #304, opened 2016-07-20
Autosave ng2

KorinaGraber:
Use some more WIP-blocked time to knock out autosave functionality. Autosave is now simply a directive that attaches to the form to provide autosave behavior. On form status changes, check the validation state and if valid, initiate an autosave cycle. The cycle begins with a debounce timer that will trigger the save itself. Keyup events anywhere on the form will reset the timer. Also, if a keyup event causes the form to become invalid, the save is canceled until validation is resolved. Since this is an autosave, no validation error is notified, as that would lead to spamming the page with toasts.

TODO: Implement ability to save the form in an invalid state with a flag set. This is used in MUSIC for 'business rule' validation, such as arrived time being after received time. We still want to save the fields, the user just can't close the SE until the error is resolved.


KorinaGraber:

PR #305, opened 2016-07-21
Export the jquery provider from the jquery utility

KorinaGraber:
This makes it available to external consumers


KorinaGraber:

PR #307, opened 2016-08-02
Rating bar ng2

KorinaGraber:
Some WIP-blocked work to implement a simple ng2 rating bar. Should have no external impact on existing package consumers.


bootstrapper/misc/miscNg2Context.ts
```diff
@@ -2,14 +2,18 @@ import { Component, ViewChild } from '@angular/core';
...
-	directives: [BusyComponent],
+	directives: [BusyComponent, RatingBarComponent, SpinnerComponent],
 })
 export class MiscNgContextBootstrapper {
 	@ViewChild('busy2') busy2: BusyComponent;
+	width: number = 100;
+	value: number = 0;
 
 	wait(): Observable<void> {
 		const subject: Subject<void> = new Subject<void>();
```
JamieBowen, line 19: Out of curiosity, do you prefer this approach over `const subject = new Subject<void>();`?  Feels like repeating yourself, and TypeScript would still implicitly type the variable as a `Subject<void>`.

JamieBowen:
:shipit:

bootstrapper/misc/miscNg2Context.ts
KorinaGraber, line 19: Good point. I think I do still like the redundancy of explicitly typing in the case of class properties, like the `width` and `value` above, but would be perfectly content to drop them for locals like this.

JamieBowen, line 19: Yeah, it doesn't seem as egregious with numbers, strings and booleans.  I still prefer to see those types explicitly declared.  I'm mainly in favor of dropping the explicit typing on complex objects where the type would appear on both sides of the equal sign.

KorinaGraber, line 19: :+1:

PR #310, opened 2016-08-03
Swapped out using 'error-message' for just alerts.

BrentWMiller:
Issue: https://renovo.myjetbrains.com/youtrack/issue/THM-42

image


KorinaGraber:
:shipit:

PR #313, opened 2016-08-08
Inputs fix

KorinaGraber:
This change was driven by the need to accommodate a case Andy ran into when migrating the status selector to angular 2. We were having issues where inputs were a little too trigger happy with firing change events. First, if you set a value model, the change event would fire twice with the initial selection. Then also, changing the value caused two change events to be fired. The second issue was causing trouble in the service times editor, and we did a temporary workaround for it there.

Also, tweaked select inputs to not fire a change event when re-selecting the currently selected value.


arknotts:
:shipit:

PR #314, opened 2016-08-08
Only emit a change as part of the setValue

KorinaGraber:
We should really only emit changes that were triggered from within the input. Also, fixes an issue we were having when an output event triggers a change to the input. This change gets picked up here and fires another output.

Temporarily violating the WIP to get this change made for Andy. Should have no QA impact anyway.


arknotts:
:shipit:

PR #316, opened 2016-08-09
Angular mocha testing

KorinaGraber:
Updated to angular rc 4. Simplified tests with a lot of dependencies using the angular test tooling to specify that we're using utility providers.

Fixed bugs with the angular 2 simple cards and long click button caught when regression testing the angular 2 components. 
In the long click button, failed long clicks would log an error to the console. Addressed by modifying the logic so the timeout catch handler fires the warning message. This is the intended purpose for why I made the timeout throw on a cancel.
In simple cards, a naming conflict with the base class was resulting in an infinite loop when the card tried to close.


KorinaGraber:
Verified in regression testing that the forms still work in the new release candidate

KorinaGraber:
The other thing we may want to look into at some point is the angular tooling for building components with their html. That might let us actually unit test the view for some key components.

JoshDSommer:
:shipit:

source/components/simpleCardList/simpleCard.ts
```diff
@@ -81,10 +81,6 @@ export class SimpleCardComponent<T> extends FormComponent implements OnInit {
 		return canClose;
 	}
 
-	saveForm(): boolean {
-		return this.boolean.toBool(this.submit());
```
JamieBowen, line 85: It looks like the boolean service is no longer being used so the import statement and DI can be removed as well.

source/components/buttons/buttonLongClick/buttonLongClick.ts
```diff
@@ -43,15 +43,11 @@ export class ButtonLongClickComponent extends ButtonAsyncComponent {
 		this.timer = this.timeoutService.setTimeout((): void => {
 			this.cleanup();
 			this.triggerAction($event);
-		}, this.duration);
+		}, this.duration).catch(() => this.warn());
 	}
```
JamieBowen, line 47: Nice improvement to eliminate the console errors

source/components/buttons/buttonLongClick/buttonLongClick.ts
```diff
@@ -43,15 +43,11 @@ export class ButtonLongClickComponent extends ButtonAsyncComponent {
...
 
 	stopAction(): void {
-		if (this.active) {
-			if (this.timer != null) {
-				this.warn();
-			}
-
+		if (this.timer) {
```
JamieBowen, line 50: Truthy FTW :+1:

source/components/simpleCardList/simpleCard.ts
KorinaGraber, line 85: :+1:

KorinaGraber, line 85: https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/316/commits/96f60d44fd376be85d1ac3e5170420c7827ab672

JamieBowen:
:shipit:

PR #317, opened 2016-08-10
Synchronized requests refactor

KorinaGraber:
This started as a spike, but I realized some valuable refactoring could be done without changing the external contract.

rxjs natively supports ignoring all but the latest request. Instead of injecting a service to handle this, we'll just convert any promises to observables and synchronize via rxjs. This removes a dependency and also allows us to support observable requests in addition to promises

Also included some bugfixes in the injections to simple card, card, and selectable card, which don't seem to be playing nicely with each other.

### DIFF
--- a/bootstrapper/cards/cardsNg2Bootstrapper.ts
+++ b/bootstrapper/cards/cardsNg2Bootstrapper.ts
@@ -1,5 +1,6 @@
 import { Component } from '@angular/core';
 import { mapValues, map, range } from 'lodash';
+import { BehaviorSubject, Observable } from 'rxjs';
 
 import { services } from 'typescript-angular-utilities';
 import __date = services.date;
@@ -59,7 +60,7 @@ export class CardsBootstrapper {
 		this.options = [1, 2, 3, 4, 5];
 
 		this.builder = cardContainerBuilder;
-		this.builder.dataSource.buildSimpleDataSource(items);
+		this.builder.dataSource.buildDataServiceDataSource<ICardItem>(() => this.wait(items));
 		this.builder.usePaging();
 		this.builder.addColumn({
 			name: 'name',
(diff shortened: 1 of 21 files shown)


bootstrapper/cards/cardsNg2Bootstrapper.ts
```diff
@@ -59,7 +60,7 @@ export class CardsBootstrapper {
 		this.options = [1, 2, 3, 4, 5];
 
 		this.builder = cardContainerBuilder;
-		this.builder.dataSource.buildSimpleDataSource(items);
+		this.builder.dataSource.buildDataServiceDataSource<ICardItem>(() => this.wait(items));
```
JoshDSommer: Could this be better accomplished by using a operator such as timeout instead of a custom wait function?

KorinaGraber: You mean something like `Observable.of(items).timeout(1000);`?

KorinaGraber: `Observable.of(items).delay(1000)` seems to work

JoshDSommer: Sweet!

KorinaGraber: https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/317/commits/8778849c91ab881be78d69ad1c189d7c8e62c9db

bootstrapper/inputs/inputsNg2Bootstrapper.ts
```diff
@@ -64,18 +64,12 @@ export class InputsBootstrapper {
...
-		return subject.asObservable();
-	}
-
 	getOptions = (): Observable<any> => {
 		return Observable.of(this.typeaheadOptions);
 	}
 
 	searchOptions = (search: string): Observable<any> => {
-		return this.wait(filter(this.typeaheadOptions, option => option.value.search(search) != -1));
+		return Observable.of(filter(this.typeaheadOptions, option => option.value.search(search) != -1)).delay(1000);
```
KorinaGraber, line 72: Note: I was going to change this to filter with rxjs, but that filters the stream itself. In this case, the stream is only ever one event, and I want to actually filter the list instead.

KorinaGraber, line 72: Could've applied this filter in a map() I guess, but that might be overkill, since this is all in the bootstrapper anyway.

JamieBowen:
:shipit:

PR #318, opened 2016-08-11
Add ability to enable/disable filter groups

arknotts:
This adds the ability to supply an input to the `rlFilterGroup` component, determining whether it is enabled or not. If not enabled, the group is not expandable and only the group label displays in the header (active option label is hidden). I've also upgraded the NG2 version and added tests.

Completes https://renovo.myjetbrains.com/youtrack/issue/MUSIC-923
CSS changes will be included in https://renovo.myjetbrains.com/youtrack/issue/MUSIC-914 (but are not required to merge)

This is a non-breaking change as it defaults to enabled if not supplied.

image

### DIFF
--- a/source/components/cardContainer/filters/filterGroup/filterGroup.directive.ng1.ts
+++ b/source/components/cardContainer/filters/filterGroup/filterGroup.directive.ng1.ts
@@ -20,7 +20,7 @@ export class FilterGroupController {
 	icon: string;
 	filterGroup: IFilterGroup;
 	source: IDataSource<any>;
-	enabled: boolean;
+	disabled: boolean;
 
 	hasIcon: boolean;
 	expanded: boolean;
@@ -29,11 +29,10 @@ export class FilterGroupController {
 	constructor(private $scope: angular.IScope) {
 		this.hasIcon = this.icon != null && this.icon !== '';
 		this.expanded = true;
-		this.enabled = true;
 	}
 
 	get headerTitle(): string {
-		if (this.enabled) {
+		if (!this.disabled) {
 			return this.filterGroup.label + ': ' + this.filterGroup.activeOption.label
 		}
 
@@ -45,7 +44,7 @@ export class FilterGroupController {
 	}
 
 	get childrenVisible(): boolean {
-		return this.expanded && this.enabled;
+		return this.expanded && !this.disabled;
 	}
 
 	selectOption(option: IFilterOption): void {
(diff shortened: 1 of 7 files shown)


source/components/cardContainer/filters/filterGroup/filterGroup.directive.ng1.ts
```diff
@@ -54,5 +68,6 @@ export let filterGroup: angular.IComponentOptions = {
 		icon: '=',
 		filterGroup: '=',
 		source: '=',
+		enabled: '<',
```
KorinaGraber: This is a little nit-picky, but it seems like it's a lot more common in our code-base to allow an option for `disabled` instead of `enabled`. That way it defaults to undefined, which is falsy. Can we just switch it for consistency's sake? Unless you have a reason/preference towards 'enabled'

KorinaGraber:
Looks good. Left one suggestion, then :shipit:

source/components/cardContainer/filters/filterGroup/filterGroup.directive.ng1.ts
arknotts: I tend to prefer `enabled` rather than `disabled` because to me it reads a lot easier. If I see something like `!disabled` or `disabled == false` I read it as "not disabled", and especially when chained with a bunch of other boolean expressions I sometimes have to do a double take on it. While something like `!enabled` or `enabled == false` reads "not enabled" which is slightly less double negative. :smile: 

That said, there's something to say for consistency in the code base so I don't mind changing it if that's how it's usually done elsewhere.

KorinaGraber: So after some discussion with @SFarageNIS, we recognized that the cases when you'd say `!disabled` or `disabled == false` are actually pretty rare. By default the component will always be enabled. So you'd only ever specify `disabled` if you specifically wanted to disable it under certain conditions. So at this point we'd like to just stick with using disabled, since it would take some migration to get `enabled` as a consistent pattern.

SFarageNIS: So the case we're generally expecting is `disabled == true`.  If we frequently checked the conditions you stated, then the solution you stated totally makes way more sense!  But...we're hopefully not doing that, since the disabled state should be the exceptional state, not the default state.  (Sam said it well, I just like typing right now)

arknotts: Makes sense, pushed a commit renaming it.

PR #319, opened 2016-08-15
Add custom placeholder to card container search card

arknotts:
This adds (or rather just exposes) the card search placeholder text to the card container input bindings so consuming pages can customize the text. This is a non-breaking change since it still falls back to the default placeholder text.

Completes https://renovo.myjetbrains.com/youtrack/issue/MUSIC-924

image

### DIFF
--- a/bootstrapper/cards/cardContainerBootstrapper.ts
+++ b/bootstrapper/cards/cardContainerBootstrapper.ts
@@ -19,6 +19,7 @@ class CardTestController {
 	builder: ICardContainerBuilder;
 	builderWithSelectFilter: ICardContainerBuilder;
 	builderWithDateFilter: ICardContainerBuilder;
+	builderWithSearch: ICardContainerBuilder;
 	options: number[];
 	selectFilter: ISelectFilter<number>;
 	dateFilter: IDateFilter;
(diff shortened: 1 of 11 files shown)


JoshDSommer:
:shipit:

KorinaGraber:
With card containers, we set most of the configuration options through the card container builder. However, in this case the search placeholder is a view concern, so I could see the argument for binding it to the view instead. But if this is the route we're going to go, you should probably add it as an input to the ng2 `selectableCardContainer` as well.

arknotts:
Yeah I noticed pretty much everything was happening through the builder. I tried setting it through the builder, but it didn't bind the value and any changes made to it in service events didn't update the text in the card container. I'd guess a watch would work but that seems dirty so I bound it in the view, unless you know of a way to do that still using the builder.

I can add it to the `selectableCardContainer` as well. Didn't realize that existed!

KorinaGraber:
Did you configure it to set the builder property on the card container in `setCardContainerProperties`?

KorinaGraber:
https://github.com/RenovoSolutions/TypeScript-Angular-Components/blob/master/source/components/cardContainer/builder/cardContainerBuilder.service.ts#L132

KorinaGraber:
And yes, in angular 2 I split up the card container into two separate components to reduce the clutter of switching views when selection is enabled.

arknotts:
Yeah, I set it within `setCardContainerProperties`, and got it to display fine. The problem is if that text needs updated after the first initialization it doesn't happen because it's not actually bound to the variable like it would be if passed in via the `Input()` (ng2) or `bindings: {...}` (ng1). It just kept its original value.

KorinaGraber:
Oh, I see. In that case the binding makes sense. The builder wouldn't include 'active' bindings, as it's designed to facilitate 'creating' the card container. (It is a builder, after all)

arknotts:
Added the same thing to the `SelectableCardContainer`. It's too bad Angular 2 doesn't support decorator inheritance as it isn't too DRY like it is now (repeating the `Input()` binding in both child and parent components) but I'm not aware of a way to get around this easily.

KorinaGraber:
What we have done sometimes in the past is define a 'inputs' object for the parent which you use to specify inputs in the @Component decorator, then reuse those in the child. That way you can just specify the property in the base class, but it's a little less clear then the @Input() decorator.
Example:
https://github.com/RenovoSolutions/TypeScript-Angular-Components/blob/master/source/components/cardContainer/card/card.ts#L18
https://github.com/RenovoSolutions/TypeScript-Angular-Components/blob/master/source/components/cardContainer/card/selectableCard.ts#L19

KorinaGraber:
:shipit:

PR #320, opened 2016-08-16
Added tests for the search stream behavior

KorinaGraber:
This is one area where my test coverage fell short the first time around, because I didn't know how to test these rxjs operators. Now that rxjs scheduling is possible, added test coverage for the search stream behavior. Essentially there are two main cases:
1) User changes the search, gets the debounce spinner, but then changes it back before the search fires. The spinner will then hide and the `distinctUntilChanged()` operator will suppress the event.
2) User changes the search and the debounce elapses. The typeahead then fires off a reload request and triggers the spinner on the request.

Needs https://github.com/RenovoSolutions/TypeScript-Angular-Utilities/pull/161


KorinaGraber:
Updated to get the necessary utilities changes

KorinaGraber:
Also filled in some holes in the card container test suite

PR #321, opened 2016-08-17
rlAsync stops displaying loading when stream complete

JoshDSommer:
For streams with more than one event, rlAsync was incorrectly showing loading as done after the first event. by switching it to only set loading to false when complete, the loading symbol is removed at the end of the event stream.

paired with @arknotts

### DIFF
--- a/source/components/busy/busy.tests.ts
+++ b/source/components/busy/busy.tests.ts
@@ -42,8 +42,14 @@ describe('busy', () => {
 			expect(busy.loading).to.be.true;
 		});
 
-		it('should finish after an event is received', (): void => {
+		it('should not finish after an event is received', (): void => {
 			stream.next(null);
+
+			expect(busy.loading).to.be.true;
+		});
+
+		it('should finish after an event is complete', (): void => {
+			stream.complete();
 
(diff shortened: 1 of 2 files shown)


KorinaGraber:
:shipit:

KorinaGraber:
What was the use case, if you don't mind my asking?

KorinaGraber:
For historical purposes, the use case was that two streams were being merged, and one event updated the UI, while another indicated the completion of the API request. Also, we suspect that using an onComplete hook will also stop the spinner when an error is received.

KorinaGraber:
We had thought about adding support for giving the busy spinner a stream of true/false events to show/hide the spinner overtime. That could still be a nice improvement, but the behavior implications are different enough that it would probably need to be a completely different method. Otherwise you'd probably run into unexpected cases where a stream fires true or false and the spinner gets shown/hidden, when it was expected to just hide after the stream completes.

JoshDSommer:
Also sync button wouldn't work with `Observable.empty()` since there is nothing to subscribe to only a completed event sent,  it would keep spinning indefinitely.

PR #322, opened 2016-08-18
Add FocusOn directive and add to card container

arknotts:
This adds an rl-focus-on directive to angular 1 and consumes it in the ng1 card container. This directive will conditionally focus an input field based on the passed in condition. If this is added directly to an input control, the input control will be focused. If not, it will search for an input control within the child controls to focus. No ng2 version was implemented because ideally ng1 will go away in the near future and in ng2 it's easier to target elements for focus without a separate directive. 

Resolves https://renovo.myjetbrains.com/youtrack/issue/MUSIC-933

### DIFF
--- a/source/behaviors/behaviors.module.ts
+++ b/source/behaviors/behaviors.module.ts
@@ -6,8 +6,9 @@ import * as autosave from './autosave/autosave.ng1';
 import * as offClick from './offClick/offClick.ng1';
 import * as popover from './popover/popover';
 import * as required from './required/required';
+import * as focusOn from './focusOn/focusOn.ng1';
 
-export { alias, autosave, offClick, popover, required };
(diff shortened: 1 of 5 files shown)


KorinaGraber:
In some ways it would be nice if focus was an action you called to tell the card container to focus the search box, rather than a flag, but that would be really hard to do in angular 1. Maybe the angular 2 card container can eventually support that.
:shipit:

KorinaGraber:
Logged https://renovo.myjetbrains.com/youtrack/issue/IDEAS-38. We don't have to take care of it now, but we'll need it before we migrate the SE list to angular 2

arknotts:
I agree, it's a little awkward having this as an input binding but seemed the only reasonable way in angular 1.

PR #323, opened 2016-08-18
Card container refactoring

KorinaGraber:
Enforce better separation of concerns within the angular 2 card container. 

1) Moved functionality for 'hydrating' column size properties to the sizeForBreakpoints service.
2) Created a sortManager service for managing the sort list for the card container.

Only effects angular 2 so this shouldn't need any QA testing. (The angular 2 card container isn't in use yet)

Depends on https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/320


source/components/cardContainer/card/headerColumn/sizeForBreakpoints.tests.ts
```diff
@@ -0,0 +1,74 @@
...
+	});
+
+	it('should convert an object of breakpoint sizes into a class string', () => {
+		const sizes = {
+			lg: 4,
+			md: 3,
+			sm: 2,
+			xs: 1,
+		};
+		expect(sizeForBreakpoints.getClass(sizes, '')).to.equal('col-xs-1 col-sm-2 col-md-3 col-lg-4 ');
```
arknotts, line 21: More curious than anything, but do we want our unit tests tightly coupled to our CSS framework? More tests always help, but it would mean that unit tests would need updating if the CSS framework changed/was updated.

source/components/cardContainer/cardContainer.html
```diff
@@ -15,7 +15,7 @@
 					</div>
 					<div class="clearfix"></div>
 				</div>
-				<div *ngIf="dataSource.dataSet | isEmpty:false">
+				<div *ngIf="hasItems">
```
arknotts, line 18: Nice! Looks like we can now unit test `hasItems` as well :+1:

source/components/cardContainer/card/headerColumn/sizeForBreakpoints.tests.ts
KorinaGraber, line 21: That's a good suggestion, but I think these are certainly still useful. Part of the purpose of unit tests is that they enforce that when you go and make a code change, the behavior still lines up with what it did before, or changes in a way that you intended (by updating the tests). Ideally, if you change the css framework you'd first update these tests to show what the classes should look like, then go update the code to match.

arknotts, line 21: Ideally I think the view would be the only piece that has any notion of column sizes, and not the code but I know the purpose of the card container is to make that whole process easier so I'm kinda nitpicking and this point. It really doesn't matter to me much; more hypothetical than anything :smiley:. There's also probably an abstraction we could use to decouple the two, but I'm sure it's not worth the overhead. I'm good with it :+1:

arknotts:
:shipit:

source/components/cardContainer/card/headerColumn/sizeForBreakpoints.tests.ts
KorinaGraber, line 21: Isn't this the abstraction though? The whole purpose of this class is to know how to turn a size configuration setting into a css class so the card container doesn't need to know how to do that

arknotts, line 21: Ah yes, you're right. :sheep: I was thinking that was part of the card container column. Now that I look at the class more closely it basically does exactly what was in my mind when I said that (mapping an abstracted column size to a css class). Except I was thinking the view would do the abstract column size -> class name conversion so it stays purely in the presentation layer.

KorinaGraber, line 21: A nice idea, but at what point does the view get too logic intensive? Particularly since views in general are un-testable or at least much more difficult to test.

arknotts, line 21: Agreed, I always opt for the dumbest views possible. I don't think a simple mapping is too intensive though, and something pretty unlikely to break outside of a bootstrap upgrade/replacement (where we would likely be doing a global search/replace anyway).

PR #324, opened 2016-08-18
Updating the MSI to use the newer style, for both ng1 and ng2.

BrentWMiller:
**Youtrack Issue: https://renovo.myjetbrains.com/youtrack/issue/THM-60**
- Added 'rl' prefixed classes to each piece of the MSI. Easier to change this way.
- Moved the error class from it's own element to being applied on the step item itself.

**Requires Theme changes to display properly, otherwise it will just show up as a list.**
**Theme PR: https://github.com/RenovoSolutions/RenovoTheme/pull/231**
### Old

image
### New

new-msi


KorinaGraber:
:shipit:

PR #325, opened 2016-08-18
Add ability to clear data source client-side.

arknotts:
This lets consuming code clear the data source without making a call back to the API.

Completes https://renovo.myjetbrains.com/youtrack/issue/MUSIC-939


KorinaGraber:
:shipit:

PR #326, opened 2016-08-19
Fix for busy if an error is thrown

JoshDSommer:
currently if an error is thrown then the busy image will continue on until the page is refreshed wich is especially problematic on async buttons.


KorinaGraber:
:shipit:

PR #328, opened 2016-08-23
Add $inject to focusOn for strict mode in RL21.

arknotts:
The bootstrapper was OK without it, but RL21 needs it so the files can be minified and bundled.

Relates to https://renovo.myjetbrains.com/youtrack/issue/MUSIC-933


KorinaGraber:
We might want to look into requiring this in the bootstrapper too. eventually. That way we won't keep running into this

PR #331, opened 2016-08-31
Add card container w/ filters example to bootstrapper

arknotts:
The filters were just in there as an example, not hooked up to a card container. This PR creates a card container and hooks them up so they will filter the data within.

Relates to https://renovo.myjetbrains.com/youtrack/issue/MUSIC-908, but doesn't contain any code right now that resolves it. This was more to make it testable in the bootstrapper.


KorinaGraber:
:shipit:

PR #336, opened 2016-09-07
Tooltip fix

KorinaGraber:
In angular 2, it's not possible to mix [] and {{}} syntax. Addressed by moving the expression to the controller. This also makes it testable.


BrentWMiller:
Nice change :shipit:

PR #338, opened 2016-09-15
Updated to Rc5 and introduce NgModules

KorinaGraber:
Updated to RC 5 and added several NgModules. ComponentsSharedModule encapsulates just the components, and can be used in any number of modules or lazy-loaded modules. ComopnentProvidersModule encapsulates the component providers, and should be used only at the application root. ComponentsModule includes both of the other two and can be used at the application root if the main application module wants to use some of the components as well.

Also added an editor config

https://renovo.myjetbrains.com/youtrack/issue/RLv21-51

Requires https://github.com/RenovoSolutions/TypeScript-Angular-Utilities/pull/164


JoshDSommer:
Instead of exporting one module containing all of these, would it be better to export individual modules, such as ButtonModule, InputsModule, CardContainerModule, etc?

KorinaGraber:
That could be nice, although I feel like both would be nice, because one of the nice things about typescript-angular-components is that it pulls together a lot of components you can use. So we could eventually make it optional to use individual modules (Button, Inputs, CardContainer, etc), or just include the full module

KorinaGraber:
Created https://renovo.myjetbrains.com/youtrack/issue/RLv21-63. I don't think this is something we necessarily have to do now, since our modules will want all of the components for the most part.

KorinaGraber:
On hold until we move to one project

PR #339, opened 2016-09-22
Signature input

KorinaGraber:
Implemented an angular 2 signature input with jSignature. After looking around, this seemed to be one of the top choices for signature components. It's nicer than signature_pad in general, and also supports sizing to the container it's put in, rather than having to set a pixel height and width. 

Dropped the old (broken) signature pad control.


KorinaGraber:
Also reproduced the issues with the previous signature pad on Firefox. This one works fine.

source/components/inputs/signature/signature.ts
```diff
@@ -0,0 +1,105 @@
...
+	selector: 'rlSignature',
+	template: require('./signature.html'),
+	inputs: validationInputs,
+	outputs: baseOutputs,
+	providers: [ComponentValidator],
+	directives: [ButtonComponent],
+})
+export class SignatureComponent extends ValidatedInputComponent<string> implements AfterViewInit, AfterViewChecked, OnChanges {
+	private _canvas: JQuery;
+	@ViewChild('canvas') set canvas(value: JQuery) {
```
arknotts: A little nitpicky, but I had to do a double take on this line. It might be more readable if the `@ViewChild` is on its own line above the property?

arknotts (approved):
Looks good to me, save for one small comment/suggestion.

source/components/inputs/signature/signature.ts
KorinaGraber: I guess I can see that for a property with logic. Normally I like to one-line these, but obviously it's not one-line anyway. 😀

arknotts: Same here usually. I do have more of a C# background than Typescript so maybe that's why 😀

PR #343, opened 2016-09-28
Wait to trigger a render until after the current execution loop

KorinaGraber:
This seems to solve an issue where the jSignature control thinks its container has no size while it's in a context that's initializing. (Such as a dialog) This forces it to wait to render until after the dialog is fully open, thus sizing it correctly.


JamieBowen:
:shipit:

PR #344, opened 2016-09-28
Form improvements

KorinaGraber:
In one of the later RCs, angular introduced functionality for programmatically marking forms as pristine or dirty, as well as for resetting the form. Previously, we were manually doing this by setting the form's internal _dirty and _pristine properties, just as a short-term workaround. Update to use angular's method instead. Also expose a method for doing a full form reset if desired.

This should have no client-facing changes, else I would've waited for an opening in the WIP first.


TheOriginalJosh:
:shipit:

JoshDSommer (approved):
:shipit:

PR #347, opened 2016-09-30
On a change, reset the control value

KorinaGraber:
For some reason I haven't been able to determine yet, the ngOnChanges hook gets fired with a change whenever the user draws on the control, so using the on-change hook at face value is causing a weird 'shift' issue. As a workaround, I'm saving the last internal change whenever the signature is updated from inside the control and comparing that against the value in the ngOnChanges hook.


arknotts:
:shipit:

PR #349, opened 2016-10-07
[Ng1] rlSelect, close multiple menus

BrentWMiller:
**Youtrack issue: https://renovo.myjetbrains.com/youtrack/issue/THM-74**

The original issue was that when you opened an rlSelect menu and then another rlSelect menu the first wouldn't close. So the menus would start to stack, or layer on top of one another.

This switches the ng1 select from using `ng-click` to using `ng-mousedown` and `ng-blur`. It also drops the need for the `rl-off-click` behavior.

These commits are a lot of trial and error. The last 4 commits are the final idea, once I figured out that `ng-focus` was breaking it and I could switch back to a toggle/close.

ng1-select-fix

### DIFF
--- a/source/components/inputs/select/select.ng1.tests.ts
+++ b/source/components/inputs/select/select.ng1.tests.ts
@@ -102,23 +102,17 @@ describe('SelectController', () => {
 			buildController();
 		});
 
-		it('should toggle the options', (): void => {
-			expect(dropdown.showOptions).to.be.undefined;
-
-			dropdown.toggle();
+		it('should open the options', (): void => {
+			dropdown.openOptions();
 
 			expect(dropdown.showOptions).to.be.true;
-
-			dropdown.toggle();
-
-			expect(dropdown.showOptions).to.be.false;
 		});
 
 		it('should close the options', (): void => {
-			dropdown.showOptions = true;
-			dropdown.close();
+			dropdown.closeOptions();
+
 			expect(dropdown.showOptions).to.be.false;
-		});
+		})
 
 		it('should do nothing if the options are already closed', (): void => {
 			dropdown.showOptions = false;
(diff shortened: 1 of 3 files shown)


source/components/inputs/select/select.ng1.ts
```diff
@@ -96,10 +96,8 @@ export class SelectController extends InputController {
 		this.showOptions = !this.showOptions;
 	}
 
-	close: { (): void } = () => {
-		if (this.showOptions) {
```
KorinaGraber, line 100: Why this change? If showOptions is already true, it shouldn't matter whether you clear it or not.

KorinaGraber (changes requested):
Question

source/components/inputs/select/select.ng1.tests.ts
```diff
@@ -117,14 +117,9 @@ describe('SelectController', () => {
...
 			dropdown.showOptions = true;
 			dropdown.close();
-			expect(dropdown.showOptions).to.be.false;
-		});
 
-		it('should do nothing if the options are already closed', (): void => {
-			dropdown.showOptions = false;
-			dropdown.close();
 			expect(dropdown.showOptions).to.be.false;
-		});
```
KorinaGraber, line 127: If the below changes, this test can come back too

source/components/inputs/select/select.ng1.ts
KorinaGraber: It looks like this more or less just happened because you were trying different things so when you came back to this it was different than when you'd started...?

BrentWMiller: It was lost while trying different things and the newer method  worked how I wanted it. Is there an advantage to the original method versus mine?

KorinaGraber: I was thinking about that after I commented. I can't really think of a reason, so we can probably leave it as-is.

KorinaGraber (approved):

source/components/inputs/select/select.ng1.ts
BrentWMiller: 👍 Okay, I just wanted to make sure I wasn't misunderstanding something.

PR #350, opened 2016-10-14
SPIKE - Step component

KorinaGraber:
Created a simple component for multi-step or tab links. The current multi-step indicator has issues with change detection because the steps are mutable. By making each step its own component, we can make the view a simple product of the inputs, which works well with OnPush change detection. 

Also, I discussed with @BrentWMiller  the possibility of making tab and MSI styling that uses very similar markup. In this component, we can support a flag that toggles the styling between a tab and MSI item, and this can then be used as a link of either type. (This will be useful for cases where we want a tab control that ties in with the url)

### DIFF
--- a/source/components/multiStepIndicator/step.component.ts
+++ b/source/components/multiStepIndicator/step.component.ts
@@ -9,4 +9,5 @@ export class StepComponent {
 	@Input() title: string;
 	@Input() link: any[] | string;
 	@Input() activeClass: string = 'active';
+	@Input() valid: boolean;
 }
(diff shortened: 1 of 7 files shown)


source/components/multiStepIndicator/step.component.ts
```diff
@@ -0,0 +1,19 @@
...
+@Component({
+	selector: 'rlStep',
+	template: require('./step.component.html'),
+	changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class StepComponent {
+	@Input() title: string;
+	@Input() link: any[] | string;
+	@Input() valid: boolean;
+	@Input() msiStyling: boolean;
```
KorinaGraber: This could probably be better named. This flag toggles the styling between MSI and tab (defaults to tab right now). Suggestions?

PR #351, opened 2016-10-19
Move initialization to $onInit

ibedard16:
Sam: After further investigation, this isn't the root problem, but it's something that should be done. With angular 2 in place, bindings aren't usually setup in general until the $onInit hook


arknotts:
:shipit:

PR #352, opened 2016-10-20
Observable validation

KorinaGraber:
Validation should be represented as observable streams. This should allow us to address issues with change detection for input validation. In particular, validation needs to be able to update when external data changes that impacts the validation. This can be done by tying a stream representing the external data into the validation stream.

Should address https://renovo.myjetbrains.com/youtrack/issue/MUSIC-1102 and https://renovo.myjetbrains.com/youtrack/issue/MUSIC-1101


arknotts:
:shipit:

PR #353, opened 2016-10-20
Ng2 label slide animations

BrentWMiller:
**Youtrack issue: https://renovo.myjetbrains.com/youtrack/issue/THM-78**

This PR sets up label animations for the ng2 versions of our form controls. It also sets up a structure for future animations to be added and used easily.

They work slightly different, if the field is focused the label will slide up now.

testing-ng-animation
testing-ng-animation_final

### DIFF
--- a/source/components/inputs/input.ts
+++ b/source/components/inputs/input.ts
@@ -1,4 +1,4 @@
-import { AfterViewInit, OnInit, EventEmitter } from '@angular/core';
+import { Component, AfterViewInit, OnInit, EventEmitter, AnimationEntryMetadata } from '@angular/core';
 import { FormControl } from '@angular/forms';
 
 import { services } from 'typescript-angular-utilities';
@@ -6,16 +6,18 @@ import __object = services.object;
 import __guid = services.guid;
 
 import { FormComponent } from '../form/form';
+import { slide } from '../../animations/animations';
 
 export const baseInputs: string[] = ['name', 'label', 'value', 'disabled', 'labelState'];
 export const baseOutputs: string[] = ['change', 'valueChange'];
+export const baseAnimations = [slide];
 
 export class InputComponent<T> implements AfterViewInit, OnInit {
 	name: string;
 	label: string = '';
 	disabled: boolean;
 	value: T;
-	labelState: string = 'hideLabel';
+	labelState: string = 'hide';
 	change: EventEmitter<T> = new EventEmitter<T>();
 	valueChange: EventEmitter<T> = this.change;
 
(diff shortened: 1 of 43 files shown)


KorinaGraber:
A nice to have would be to hide the placeholder when the label slides up. (Documenting this from our slack convo)

source/components/inputs/input.ts
```diff
@@ -36,6 +39,10 @@ export class InputComponent<T> implements AfterViewInit, OnInit {
 		if (this.object.isNullOrEmpty(this.name)) {
 			this.name = this.inputType + '-' + this.guid.random();
 		}
+
+		if (!this.object.isNullOrEmpty(this.value)) {
```
KorinaGraber: We decided to move away from this. Instead you can just use truthy here, as in `if (this.value)`

source/components/inputs/input.ts
```diff
@@ -62,4 +69,14 @@ export class InputComponent<T> implements AfterViewInit, OnInit {
 			this.change.emit(value);
 		}
 	}
+
+	showLabel(): string {
+		return this.labelState = slide.show;
+	}
+
+	hideLabel(): string {
+		return this.object.isNullOrEmpty(this.value)
```
KorinaGraber: And here this would just be `!this.value`, or `this.value` if you reverse the two cases below

source/components/inputs/input.ts
```diff
@@ -62,4 +69,14 @@ export class InputComponent<T> implements AfterViewInit, OnInit {
 			this.change.emit(value);
 		}
 	}
+
+	showLabel(): string {
+		return this.labelState = slide.show;
+	}
+
+	hideLabel(): string {
```
KorinaGraber: Is there a better way we could name this, since you're only conditionally hiding the label? It seems strange to call `hideLabel()` and then have it just call `showLabel()` instead. Thoughts?

KorinaGraber (changes requested):
This looks really good! I just had a few comments about code styles and naming.

BrentWMiller:
Updated with requested changes and fixed conflicts.

BrentWMiller:
Looking at hiding the placeholder when label slides up.

KorinaGraber (approved):

BrentWMiller:
Updated for hiding placeholder when focusing.

source/components/inputs/input.ts
```diff
@@ -1,20 +1,24 @@
...
 
 import { services } from 'typescript-angular-utilities';
 import __object = services.object;
 import __guid = services.guid;
 
 import { FormComponent } from '../form/form';
+import { slide } from '../../animations/index';
 
-export const baseInputs: string[] = ['name', 'label', 'value', 'disabled'];
+export const baseInputs: string[] = ['name', 'label', 'value', 'disabled', 'labelState', 'hidePlaceholder'];
```
KorinaGraber: I don't think these need to go in the inputs array since these are just private properties to the input class

BrentWMiller:
testing-ng-animation_final_hidepalceholder

BrentWMiller:
https://github.com/RenovoSolutions/RenovoTheme/pull/243

Theme change for hiding the place holder

PR #354, opened 2016-10-20
Autosave fix

KorinaGraber:
Don't autosave unless the form is dirty. Otherwise, we'll have random requests firing off at different times.


arknotts:
:shipit:

PR #356, opened 2016-10-24
Observable card container

KorinaGraber:
Initial groundwork for a new angular 2 card container entirely using observables

The key here is that using observables makes change detection much easier in angular 2, and allows OnPush change detection. By representing the data sets as a stream, we can represent all filtering, paging, sorting, etc as maps against the data sets and everything should automatically update when the data set changes. For the moment, only search filters are supported, so we'll want to implement more filters as observable filters before this will be really usable.

Also, the only data source currently in the new system is an observable data source. This can take the place of the old simple or data service data sources, as you can simply give it `Observable.of(array)` or the request object.

Also need to reimplement the selectable card container behavior.

Currently the angular 2 card container isn't in use, so this should have no user facing impact. We'll just need to reimplement the parts we need before we actually put this in use


arknotts:
:shipit:

PR #357, opened 2016-10-24
Upgraded userRating to inherit from validationInput instead of the ba…

KorinaGraber:
…se input

This allows us to apply validation to a user rating control, such as rlRequired. (Not much reason you would apply custom validation)

Also added a test fix for the autosave directive. Forgot to update the tests when we made the change to require the form to be dirty.


arknotts:
:shipit:

PR #358, opened 2016-10-24
Change select to just take a simple array instead of an observable

KorinaGraber:
Consumers can always pipe to async if they need an observable. We'll just show the busy spinner whenever the array is null. (Pass an empty array to show an 'empty' dropdown instead)
@jsommer made the suggestion.


JoshDSommer:
:shipit:

PR #360, opened 2016-10-26
Step Component, MSI fix

BrentWMiller:
**Youtrack issue: https://renovo.myjetbrains.com/youtrack/issue/THM-80**
**Theme PR, with the needed changes: https://github.com/RenovoSolutions/RenovoTheme/pull/244**

Updated the step component to work with the MSI styling _(tab styling in the future)_. Since they both use flexbox to lay out their steps.

This change will use the host element as the 'list' item. So that the flexbox styling can be properly inherited.

These means we need to set classes on the host element instead.

updated_step

### DIFF
--- a/source/components/multiStepIndicator/step.component.ts
+++ b/source/components/multiStepIndicator/step.component.ts
@@ -9,18 +9,35 @@ export class StepComponent {
 	@Input() title: string;
 	@Input() link: any[] | string;
 	@Input() valid: boolean;
+	@Input() current: boolean;
 	@Input() useMsiStyling: boolean = false;
 
+	// Default classes
+	@HostBinding('class.active') hasActive: boolean = true;
+
+	// Conditional classes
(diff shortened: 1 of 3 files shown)


source/components/multiStepIndicator/step.component.ts
```diff
@@ -9,11 +9,18 @@ export class StepComponent {
 	@Input() title: string;
 	@Input() link: any[] | string;
 	@Input() valid: boolean;
-	@Input() useMsiStyling: boolean;
+	@Input() useMsiStyling: boolean = false;
 
-	get stepClass(): string {
+	@HostBinding('class.rl-multi-step-item') useMsi: boolean = false;
+	@HostBinding('class.rl-tab-item') useTab: boolean = false;
```
KorinaGraber: Ah, there you go. Fixes the issue of having an unstyled rlStep tag. Good fix

KorinaGraber (approved):

PR #361, opened 2016-10-27
Add validation check for invalid date

ibedard16:
Right now, we currently have no validation in the date input to assert the date is valid. This pr makes it easier for inputs to specify their own internal validators, and also makes the Date Controller check if a date is valid.


arknotts:
I like the idea (and implementation looks good), but off hand I can't think of any other components that might want their own internal validator. Assuming this is the case, we could probably simplify this and have the dateTime component directly add the validator to `super.validators` array and remove the `internalValidators` variable. Just a thought.

KorinaGraber (approved):

PR #364, opened 2016-11-04
Managing the subscriber

KorinaGraber:
As we've recently discovered, in rxjs, there should only ever be one subscriber to an http request. This is challenging because the component that initiates the action needs to know when it completes, so it can resolve functionality against it. This means that unless we added some sort of 'done()' functionality (which would be a little clunky), the initiating entity needs to trigger the final subscription.

I've identified four 'trigger points' in components. These are the async button, the submit button, the autosave directive, and the form `submit()` function (which differs from `submitAndWait()` in that it doesn't give you the observable back). For any other use cases, the consumer will most likely need to manage their own subscription. What is telling about these cases is that the consumer gives the component a 'callback' that is then called, and an observable is returned. So we can recognize the pattern that if the component wants an observable back, most likely he will be subscribing to it himself.

Corrected other usages, like `busy` and `autosaveAction` to use `.do()` instead of subscribing. These cases don't ever need to initiate anything, they just act on (or wait on) the result of the action.


KorinaGraber:
Pair reviewed with Jeff

PR #368, opened 2016-11-09
[Components] MSI and Tabs use same markup

BrentWMiller:
**Youtrack issue: https://renovo.myjetbrains.com/youtrack/issue/THM-89**

**Requires a Theme and Core change:**
* **Theme PR: https://github.com/RenovoSolutions/RenovoTheme/pull/248**
* **Core PR: https://github.com/RenovoSolutions/RL21.Application.Core/pull/659**

Altering the markup for rlStep and Tabset so they use the same 'step' markup. This way we can just apply a single class to the wrapping div and it will set the style differences for each.

Removed methods/tests related to `[useMsiStyling]=""` as it won't be needed anymore. We can now just set the single wrapping class, instead of setting this attribute for each step.

### DIFF
--- a/source/components/multiStepIndicator/step.component.html
+++ b/source/components/multiStepIndicator/step.component.html
@@ -1,5 +1,5 @@
-<div class="rl-item-wrap"
+<div class="rl-step-wrap"
 		[routerLink]="link"
 		routerLinkActive="current">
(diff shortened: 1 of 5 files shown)


KorinaGraber (approved):
💥

PR #369, opened 2016-12-02
Focus inputs

KorinaGraber:
Added the ability to focus textboxes, typeaheads, spinners, and datetimes. These are needed for the new importer. 

Paired with Jeff on the implementation.

Also updated waitOn in the autosave action service and busy component to use generic typing. If you pass an observable through waitOn and then subscribe to the result, it should have the same typings as before, rather than reverting to `any`


arknotts:
:shipit:

arknotts:
:shipit:

arknotts:
:shipit:

jtest15:
Used in this YT issue: https://renovo.myjetbrains.com/youtrack/issue/RLv2-1837

PR #370, opened 2016-12-07
tweaks to the busy component and its handling of observables

jbeck8176:
Added functionality to handle live non completing streams. Also broke the main method down into a few specific public methods and modified the main method to use them.

### DIFF
--- a/source/components/busy/busy.tests.ts
+++ b/source/components/busy/busy.tests.ts
@@ -1,13 +1,12 @@
 import { Subject, Observable } from 'rxjs';
 
 import { BusyComponent } from './busy';
-import { AsyncHelper } from '../../services/async/async.service';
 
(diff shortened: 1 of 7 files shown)


KorinaGraber (commented):
My only concern with this is that by changing the default to `next` instead of `complete`, we might break some current instances. For example, I think the return reasons component might return an empty observable to the wait on, which never fires the `next` handler.

KorinaGraber:
Couldn't we just default to `complete` to keep the behavior the same for now? We can always switch to use the others as needed

KorinaGraber (approved):

KorinaGraber:
:shipit:

KorinaGraber (approved):

KorinaGraber (approved):

PR #372, opened 2016-12-12
Set the time zone before it is needed

ibedard16:
If this value was unset, it would eventually throw an error when a function uses it just a few lines below

### DIFF
--- a/source/components/inputs/dateTime/dateTime.ng1.ts
+++ b/source/components/inputs/dateTime/dateTime.ng1.ts
@@ -117,6 +117,7 @@ export class DateTimeController extends InputController {
 		});
 
 		this.ngModel.$parsers.push((value: string): moment.Moment => {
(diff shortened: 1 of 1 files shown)


source/components/inputs/dateTime/dateTime.ng1.ts
```diff
@@ -117,6 +117,7 @@ export class DateTimeController extends InputController {
 		});
 
 		this.ngModel.$parsers.push((value: string): moment.Moment => {
+			this.timezone = __timezone.timezoneService.currentTimezone;
```
KorinaGraber: I feel like you only want to do this if the timezone is currently null. In 'most' cases, this should get set in the formatters long before this

KorinaGraber (changes requested):
One suggestion

KorinaGraber (approved):

PR #374, opened 2016-12-22
UpgradeModule

KorinaGraber:
Upgrade the components bootstrapper to use the latest technique for running an ng-upgrade. Instead of using an upgrade adapter, the application is now bootstrapped as an angular 2 application, with an `UpgradeModule` imported into the ngModule to give us access to upgrade angular 1 components and services. A side benefit is that upgrades and downgrades are now performed through static methods from `angular/upgrade/static`, and we no longer need to 

This is primarily a spike to determine how this could work for the core application. The new upgrade technique is a prerequisite for AoT compliation and lazy-loading with ngUpgrade.

**Don't merge yet**
This should have no user-facing changes in the core module, but it will require us to go to angular 2.3.0. (At least angular 2.2.0 was required for the new UpgradeModule approach)


KorinaGraber:
Tests are now functional! 🎉

PR #376, opened 2016-12-30
Fix and organize tab component

jbeck8176:


KorinaGraber:
I was thinking we would like to do this using TemplateRefs, so we can consume it the same way we consume other templated components, like simple card and card container. Also, this would give us full templating support in the header, instead of just the raw html. The drawback would be that it sounded like you had run into some issues the ngTemplateOutlet was causing in the typeahead list component. Thoughts?

KorinaGraber:
Per @jbeck8176, using template refs isn't going to work in the header because of issues with the way the tab, tabset, and tab template directives interact. The alternatives seem to be to leave it as-is or have two different styles in the tabset for defining the header content vs the other content. With this in mind, we decided to leave it as-is for now.

KorinaGraber (approved):

jbeck8176:
It was a good suggestion tho, I liked your thinking!

BrentWMiller:
Youtrack: https://renovo.myjetbrains.com/youtrack/issue/MUSIC-1255

PR #377, opened 2016-12-30
Ng module spike

BrentWMiller:
**Youtrack issue: https://renovo.myjetbrains.com/youtrack/issue/IDEAS-65**
**Paired with @TheOriginalJosh**

This just touches the button components right now.

 This will eventually help with AoT and tree-shaking.

### DIFF
--- a/source/components/busy/busy.module.ts
+++ b/source/components/busy/busy.module.ts
@@ -1,0 +1,14 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+
+import { BusyComponent } from './busy';
+
+@NgModule({
+	exports: [
+		BusyComponent
+	],
+	declarations: [
+		BusyComponent
+	],
+})
+export class BusyModule { }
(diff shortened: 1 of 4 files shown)


BrentWMiller:
Tests don't currently run: `Disconnected (1 times), because no message in 10000 ms.`

KorinaGraber:
I like this change. We'll have to figure out why the tests aren't running so we can get this working.

FYI, it's possible this might conflict with https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/374, although it hopefully won't

KorinaGraber (changes requested):
Marking as needing changes until the tests are running. That way we don't accidentally merge it

BrentWMiller:
Looks like it will conflict with #374, but shouldn't be too hard to fix. I think they both touch `ui.module.ts`.

KorinaGraber (approved):

KorinaGraber:
Needs YT issue

BrentWMiller:
YT issue added.

PR #391, opened 2017-02-13
Changes for service MSI upgrade

jbeck8176:
https://renovo.myjetbrains.com/youtrack/issue/MUSIC-1247

### DIFF
--- a/source/components/cardContainer/cardContainer.html
+++ b/source/components/cardContainer/cardContainer.html
@@ -26,7 +26,8 @@
 				<div *ngIf="hasItems$ | async">
 					<div class="card-item-repeat" *ngFor="let card of dataSource.dataSet$ | async">
 						<rlCard [item]="card"
-								[save]="save"></rlCard>
+								[save]="save"
+								[permanentFooter]="permanentFooters"></rlCard>
 					</div>
 				</div>
 
@@ -34,7 +35,7 @@
 					<rlBusy [loading]="dataSource.loadingDataSet$ | async" size="2x"></rlBusy>
 				</div>
 
-				<div class="card-container-footer" *ngIf="!(dataSource.isEmpty$ | async)">
+				<div class="card-container-footer">
 					<rlContainerFooter [footer]="containerFooter"></rlContainerFooter>
 				</div>
 			</div>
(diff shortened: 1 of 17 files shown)


source/components/cardContainer/cardContainer.html
```diff
@@ -26,15 +26,16 @@
...
+								[permanentFooter]="permanentFooters"></rlCard>
 					</div>
 				</div>
 
 				<div>
 					<rlBusy [loading]="dataSource.loadingDataSet$ | async" size="2x"></rlBusy>
 				</div>
 
-				<div class="card-container-footer" *ngIf="!(dataSource.isEmpty$ | async)">
+				<div class="card-container-footer">
```
KorinaGraber: Shouldn't this still be conditional by the current condition *or* the permanent footer property? (`*ngIf="!(dataSource.isEmpty$ | async) || permanentFooter"`)

KorinaGraber (changes requested):
One question about the changes to cardContainer.html. Otherwise, looks good.

source/components/cardContainer/cardContainer.html
KorinaGraber: Ah, I thought this was in the card...

KorinaGraber: Why did you do this change at all? The behavior we were implementing here is that the footer (paging, etc) won't show up at all if there is no data, and you'll just get the 'no data to show' message

jbeck8176: I believe I took this out because it stopped footers from showing when data was added without using the data set. Tomorrow let's look at the possible missing functionality to make sure both conditions are still being met.

KorinaGraber: Ooh, okay. So is it a problem specific to the simple data source?

jbeck8176: I think, there was a lot of changes made to that component that didn't stay, It might just be something I missed when rolling back. Im looking at adding it back in now to see if it breaks the functionality I was making the change for.

jbeck8176: I added this back in and it didn't affect the changes I had made. Great catch Sam!

KorinaGraber (approved):

source/components/cardContainer/cardContainer.html
KorinaGraber: Thanks

PR #394, opened 2017-02-17
added event emitter to emit dialog open status

jbeck8176:
I needed a way to tell if the dialog was open or closed. I added a boolean event emitter to output the change in dialog status

### DIFF
--- a/source/components/dialog/dialog.ts
+++ b/source/components/dialog/dialog.ts
@@ -1,5 +1,5 @@
 import { Observable } from 'rxjs';
-import { Component, Input, ContentChild, forwardRef } from '@angular/core';
+import { Component, OnInit, Input, Output, ContentChild, forwardRef, EventEmitter } from '@angular/core';
 
 import { services } from 'typescript-angular-utilities';
 import __notification = services.notification;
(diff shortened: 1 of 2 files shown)


JoshDSommer (approved):

source/components/dialog/dialog.ts
```diff
@@ -50,15 +55,18 @@ export class DialogComponent extends FormComponent {
...
 
 	close(): void {
 		this.dialogRoot.closeDialog.next(null);
+		this.dialogOpenStatus.emit(false);
 	}
 
 	dismiss(): void {
 		this.dialogRoot.dismissing = true;
 		this.dialogRoot.closeDialog.next(null);
+		this.dialogOpenStatus.emit(false);
```
KorinaGraber, line 69: What if we subscribed to `dialogRoot.closeDialog` instead? I'm thinking it's possible for close events to come from outside of this component. (Such as clicking outside the dialog, which will originate from the dialog outlet component.

source/components/dialog/dialog.ts
```diff
@@ -40,6 +41,10 @@ export class DialogComponent extends FormComponent {
 		this.dialogRoot = dialogRoot;
 	}
 
+	ngOnInit() {
+		this.dialogOpenStatus.emit(false);
```
KorinaGraber, line 45: Can you explain the naming?

KorinaGraber (changes requested):
Thoughts about closing dialogs?

source/components/dialog/dialog.ts
jbeck8176: the event emitter is referencing the dialog window itsself. Its a boolean representation of the openness of the window, and its a current status signifying it will and or can change. dialogOpenStatus

jbeck8176:
After messing with the stability and actual viability of this feature I decided to abort the mission

PR #395, opened 2017-02-21
Added an example of a dropdown and typeahead in a dialog

KorinaGraber:
This is causing an issue where clicking on the scrollbar in the dropdown/typeahead closes the dropdown. By reproducing it here we can try out scenarios for fixing it.


BrentWMiller (commented):
Looks good to me, gives us something to test against.

PR #397, opened 2017-02-21
Added support for setting an offClick directive as inactive

KorinaGraber:
The directive will watch for ngOnChanges and add and remove the event listener as needed. This is important because we might be going back to using off click for closing the popout list in our select and typeahead dropdowns. Without this change, that would mean that every dropdown and typeahead adds a document event to the page. Instead we can bind to offClickActive to only turn on the event listener when the dropdown is open, which will restrict us to one document listener at a time. (Only one dropdown should be able to open at a time)

### DIFF
--- a/source/behaviors/offClick/offClick.ts
+++ b/source/behaviors/offClick/offClick.ts
@@ -1,4 +1,4 @@
-import { Directive, Host, Output, EventEmitter, OnInit, OnDestroy } from '@angular/core';
+import { Directive, Host, Input, Output, EventEmitter, OnInit, OnDestroy, SimpleChanges } from '@angular/core';
 
 import { services } from 'typescript-angular-utilities';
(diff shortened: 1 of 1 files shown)


source/behaviors/offClick/offClick.ts
```diff
@@ -29,16 +30,36 @@ export class OffClickDirective implements OnInit, OnDestroy {
...
+			if (changes['active'].currentValue) {
+				this.addListener();
+			} else {
+				this.removeListener();
+			}
+		}
+	}
+
+	addListener(): void {
+		document.addEventListener('click', this.listener);
```
JoshDSommer, line 51: Same as https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/398

JoshDSommer (dismissed):

PR #398, opened 2017-02-21
Dropdowns click off to close

KorinaGraber:
Don't use focus and blur as the open/close triggers for the list. This was causing issues with the list closing when the user tried to click on the scroll bar for the list. Instead we'll go back to using an off click trigger to close the list. The list will also close on selection. Also, the user can now toggle the list by clicking on the input itself, so they can open and close it as needed by that method. This should actually fix the gimicky behavior we were occasionally seeing in the importer, when somehow focus was lost but the list was still visible. The input no longer needs to be focused in order to select an item from the list, as long as the list is open.

**Slight feature regression:** tab will no longer select the currently highlighted item in the list. I tried to add a keyup.tab event listener, but it didn't seem to work. If we want to get this behavior back, we'll have to look at how angular does tab events.

Fixes https://renovo.myjetbrains.com/youtrack/issue/RLv21-123 for angular 2 dropdowns and typeaheads

### DIFF
--- a/source/behaviors/offClick/offClick.ts
+++ b/source/behaviors/offClick/offClick.ts
@@ -15,7 +15,17 @@ export interface IOffClickEvent extends MouseEvent {
 })
 export class OffClickDirective implements OnInit, OnDestroy {
 	@Output('offClick') offClick: EventEmitter<any> = new EventEmitter();
-	@Input('offClickActive') active: boolean = true;
+
+	private _active: boolean = true;
+	@Input('offClickActive') set active(value: boolean)
+	{
+		if (value) {
+			this.addListener();
+		} else {
+			this.removeListener();
+		}
+		this._active = value;
+	}
 
 	listener: { ($event: MouseEvent): void } = ($event: IOffClickEvent) => {
 		if ($event.rlEventIdentifier !== this.identifier) {
@@ -30,20 +40,10 @@ export class OffClickDirective implements OnInit, OnDestroy {
 	}
 
 	ngOnInit() {
-		if (this.active) {
+		if (this._active) {
 			setTimeout(() => {
 				this.addListener();
 			});
-		}
-	}
-
-	ngOnChanges(changes: SimpleChanges): void {
-		if (changes['active']) {
-			if (changes['active'].currentValue) {
-				this.addListener();
-			} else {
-				this.removeListener();
-			}
 		}
 	}
 
--- a/source/behaviors/offClick/offClick.ts
+++ b/source/behaviors/offClick/offClick.ts
@@ -1,4 +1,4 @@
-import { Directive, Host, Output, EventEmitter, OnInit, OnDestroy } from '@angular/core';
+import { Directive, Host, Input, Output, EventEmitter, OnInit, OnDestroy, SimpleChanges } from '@angular/core';
 
 import { services } from 'typescript-angular-utilities';
 import __guid = services.guid;
@@ -15,6 +15,7 @@ export interface IOffClickEvent extends MouseEvent {
 })
 export class OffClickDirective implements OnInit, OnDestroy {
 	@Output('offClick') offClick: EventEmitter<any> = new EventEmitter();
+	@Input('offClickActive') active: boolean = true;
 
 	listener: { ($event: MouseEvent): void } = ($event: IOffClickEvent) => {
 		if ($event.rlEventIdentifier !== this.identifier) {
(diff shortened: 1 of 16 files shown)


KorinaGraber:
Includes https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/397

BrentWMiller (approved):

source/behaviors/offClick/offClick.ts
```diff
@@ -29,16 +30,36 @@ export class OffClickDirective implements OnInit, OnDestroy {
...
+			if (changes['active'].currentValue) {
+				this.addListener();
+			} else {
+				this.removeListener();
+			}
+		}
+	}
+
+	addListener(): void {
+		document.addEventListener('click', this.listener);
```
JoshDSommer: Could we change this to use an abstract class and a value-provider, so that document is injected much like we do window else where?

source/behaviors/offClick/offClick.ts
```diff
@@ -29,16 +30,36 @@ export class OffClickDirective implements OnInit, OnDestroy {
...
-		}, 0);
+		if (this.active) {
+			setTimeout(() => {
+				this.addListener();
+			});
+		}
 	}
 
-	ngOnDestroy() {
+	ngOnChanges(changes: SimpleChanges): void {
```
jbeck8176: You don't have hook into change detection like this. Angular does this already. You can add a setter to the input:
`@Input('active')
	set active(value: boolean) { 
		//handle the listener stuff here
	}`

JoshDSommer (changes requested):

jbeck8176 (changes requested):

source/behaviors/offClick/offClick.ts
KorinaGraber: The main reason I didn't do that was because I wasn't sure how that would work with setting a default value. I wanted this to default to true so if you didn't set active at all, the directive would still work.

jbeck8176: You should be able to set a default on a private property and add a getter to use that. I suggest this because from what I gathered it circumvents some of angular's targeted change detection.

source/behaviors/offClick/offClick.ts
KorinaGraber: Yeah, that's probably a good idea. We don't have a provider for the native document yet.

source/behaviors/offClick/offClick.ts
JoshDSommer: I'll chime in with you shouldn't be using a boolean as a tri-state. it's either false or true. null should never be a valid value.

source/behaviors/offClick/offClick.ts
jbeck8176: 👍

source/behaviors/offClick/offClick.ts
KorinaGraber: That'll work I guess. I also realized I forgot to clean up this implementation anyway, to get rid of the string typing. Usually when we do use ngOnChanges we create an interface so we can get strong typing on the changes object.

jbeck8176: 👍  I have a feeling we are gonna get some nice usage out if this, cleaner the better!

KorinaGraber: Actually @TheOriginalJosh it's not a null tri-state. That statement is checking to see if the 'changes' object has an 'active' change on it. And if there is an active change, we check for the boolean value. This is a must if we're using ngOnChanges at all, or else we'll just get null refs if it ever gets called when active didn't change.

KorinaGraber: Either way, it's a moot point given Jeff's suggestion

KorinaGraber: https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/398/commits/a5eb19c5e0e546dbbb78478c5125bd82ff7ae87d

source/behaviors/offClick/offClick.ts
KorinaGraber: https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/398/commits/ce2114106853cbed28da8deccc727d6ede2644ab

jbeck8176 (approved):

JoshDSommer (approved):

PR #401, opened 2017-02-22
Validation changes and additions

jbeck8176:
-added naming to forms when adding to parent as child controls
-added recluse form. To be used when needed to break the nesting of rlForm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-angular-components",
-  "version": "4.2.1",
+  "version": "4.2.2",
   "description": "Reusable responsive angular components",
   "author": "Renovo Development Team",
   "keywords": [
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "typescript-angular-components",
-  "version": "4.1.7",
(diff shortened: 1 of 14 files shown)


source/components/form/form.ts
```diff
@@ -76,7 +78,7 @@ export class FormComponent {
 	}
 
 	validate(): boolean {
-		return this.form.valid;
+		return !this.form.invalid;
```
KorinaGraber: Were you going to revert this since `pending` is actually another state that's not exactly valid?

KorinaGraber (changes requested):

source/components/form/form.ts
jbeck8176: Missed that one, got the other tho

KorinaGraber (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "typescript-angular-components",
-  "version": "4.2.0",
+  "version": "4.2.1",
```
BrentWMiller: Just leaving a note that this version will need updated again.

PR #405, opened 2017-03-01
Allow the consumer to override the default debounce

KorinaGraber:
On one particular page, we're getting a request to extend the debounce time before performing a search. Instead of necessarily changing this across the board, I'm just going to give us the ability to pass in a custom debounce time. If present, this should always be used.


JamieBowen (approved):

arknotts:
Looks good. 👍 A test that ensures the passed debounce time takes priority over the default would be nice too.

KorinaGraber:
Added test coverage

PR #412, opened 2017-03-14
Ng2 message log spike

KorinaGraber:
Started spiking some implementation for an angular 2 message log. We'll eventually need this for project Saturn. In the meantime, decided to leave it at this point since it's not part of the first 'MVP' release.

The idea was to use what I learned from the ngx-simple-card-container implementation to make the message log more flexible and maintainable for the future. The core log component will simply take a set of properties and display a list of messages, plus the forward and top buttons, and emit a series of events. Also, since our current message log in service events has custom templates depending on the message type, I decided to leave off a default template altogether for the moment. The consumer can presumably ngSwitch here to show something different for each message type.


KorinaGraber:
There was some debate about whether we should just put this in or private application code, since it doesn't seem as useful for open source consumers, as it's not fully featured. I'm gonna go ahead and merge it for now, and if we decide to go that route it shouldn't be too hard to delete this and copy it to another repository.

PR #415, opened 2017-03-16
Spinner default to being blank

BrentWMiller:
**YouTrack issue: https://renovo.myjetbrains.com/youtrack/issue/MUSIC-1302**
> Make labor and downtime inputs default cursor to the front of the time input and not at the end

> Acceptance Criteria:
> - The input needs to default to the front of the number input on the left side of the decimal number so that if the user clicks in the time entry area they can just type 1 and it will display it as 1.00 without having to move the cursor. 

This just removes the default of '0.00'. So the user doesn't have to backspace/delete anything. They can just type the number they want. Pretty much just a convenience change.

### Old
spinner_oooold

### New
spinner_neeeeew

### DIFF
--- a/source/components/inputs/spinner/spinner.ts
+++ b/source/components/inputs/spinner/spinner.ts
@@ -70,7 +70,7 @@ export class SpinnerComponent extends ValidatedInputComponent<number> implements
 		}
 
 		super.ngAfterViewInit();
-		this.value = this.value || 0;
+		this.value = this.value;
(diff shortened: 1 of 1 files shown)


KorinaGraber (approved):
I'd be curious to see if there are any implications of this. I was thinking that I defaulted the value to 0 for a particular reason, but maybe it's not relevant anymore. As long as it works, I'm fine with it.

PR #416, opened 2017-03-21
version bump

JoshDSommer:
needs to have a version of utilities that doesn't include .ts file or willl cause build erros for our cli applications.

 https://github.com/RenovoSolutions/TypeScript-Angular-Utilities/pull/184


KorinaGraber (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "typescript-angular-components",
-  "version": "4.2.11",
+  "version": "4.2.12",
```
BrentWMiller: We're at 4.3.0 now.

PR #424, opened 2018-01-03
Fix for opening a dialog from a dialog

KorinaGraber:
When one dialog wants to open from another, currently they both get closed immediately, since the second dialog opens and then the first one immediately closes. I tried adding a timeout or changing the order so the old dialog closes first, but this only seems to result in two overlays. Instead, we can actually fix this by disabling close events from a dialog unless it is the currently active dialog. If a second dialog opens up, the first dialog gets marked as inactive and close events are ignored from it. This way, the dialog outlet template simply gets replaced and the dialog outlet doesn't need to be closed and reopened.

### DIFF
--- a/source/components/dialog/dialog.ts
+++ b/source/components/dialog/dialog.ts
@@ -1,4 +1,5 @@
 import { Observable } from 'rxjs';
+import { Subject } from 'rxjs/Subject';
 import { Component, Input, ContentChild, forwardRef } from '@angular/core';
 
 import { services } from 'typescript-angular-utilities';
@@ -31,6 +32,7 @@ export class DialogComponent extends FormComponent {
 	@ContentChild(DialogFooterTemplate) footer: DialogFooterTemplate;
 
 	dialogRoot: DialogRootService;
+	isActive: boolean;
 
 	constructor(notification: __notification.NotificationService
 		, asyncHelper: AsyncHelper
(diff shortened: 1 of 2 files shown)


KorinaGraber:
Trying a different solution. I think this might only be needed for prompt dialogs.

source/components/dialog/dialog.ts
```diff
@@ -52,18 +54,35 @@ export class DialogComponent extends FormComponent {
 			submitAndClose: () => this.submitAndClose(),
 			size: this.size,
 		});
+		const inactive$ = new Subject()
+		this.dialogRoot.openDialog.takeUntil(inactive$).subscribe(() => {
```
JoshDSommer, line 58: I don't quite follow the logic behind these, if the dialogRoot. openDialog or closeDialog observables emit a value, then inactive$ emits a value and it completes the source. Could we just remove the inactive$ observable and replace  `takeUntil` with `take(1)`?  Or am i missing something else? I haven't used takeUntil much so I'm trying to understand it better.

KorinaGraber: I think the only think takeUntil was doing here instead of take(1) was just that it only takes the first of either of these streams. Anyway it's a moot point since I closed the PR and am trying a different approach

KorinaGraber: Replaced by https://github.com/RenovoSolutions/TypeScript-Angular-Components/pull/425

JoshDSommer: Oh IC, since inactive$ was emitting it would close both streams after one of them had emitted a value. 

As an alternative, you could have probably used the `Observable.race`  either way, the replacement is way better 👍

PR #428, opened 2018-08-13
added touch start and touch end directive to NG1 long click button.

jbeck8176:
Long click button wasn't working on mobile/touch screen in rl2. This is because touch events are fired instead of click events when browsers sense touch screens

### DIFF
--- a/source/components/buttons/buttonLongClick/buttonLongClick.ng1.ts
+++ b/source/components/buttons/buttonLongClick/buttonLongClick.ng1.ts
@@ -86,4 +86,27 @@ let longClickButton: angular.IComponentOptions = buildButton({
 
 angular.module(moduleName, [downgrade.moduleName, promiseModuleName])
 	.component(componentName, longClickButton)
-	.controller(controllerName, LongClickButtonController);
+	.controller(controllerName, LongClickButtonController)
+	.directive('myTouchstart', [function () {
+		return function(scope, element, attr) {
+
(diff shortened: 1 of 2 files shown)


source/components/buttons/buttonLongClick/buttonLongClick.ng1.ts
```diff
@@ -86,4 +86,27 @@ let longClickButton: angular.IComponentOptions = buildButton({
 
 angular.module(moduleName, [downgrade.moduleName, promiseModuleName])
 	.component(componentName, longClickButton)
-	.controller(controllerName, LongClickButtonController);
+	.controller(controllerName, LongClickButtonController)
+	.directive('myTouchstart', [function () {
```
ibedard16, line 90: The `my` prefix makes me a little bit uncomfortable. Nothing worth rejecting over though, since the directive is scoped to the long-click component.

ibedard16 (approved):

JoshDSommer:
Why do you need the custom directives? Could you just apply to the native touchstart and touchend events, I believe I've done it this way before.

Cheers!

jbeck8176:
@TheOriginalJosh There is no such thing as native touchstart and touchend events in angularJS

JoshDSommer:
@jbeck8176  Natives as in native to the browser 
https://developer.mozilla.org/en-US/docs/Web/Events/touchstart

